PR #34, opened 2026-05-04
Game Setup Phase: Player class

### LINKED ISSUE #14: Game setup phase: Player class

### LINKED ISSUE #85: One turn of the game: Player class
Fields

private final String name
private final List<Card> hand

Methods

public Player(String name)
public List<Card> getHand()
public String getName()
public void addCardToHand(Card card)
public void toggleSelectedHandCardAt(int handCardIndex)
public void removeCardFromhand(Card card)
public void deselectHandCards()

dongjoo224:


docs/bva/Player.md
```diff
@@ -0,0 +1,38 @@
+# BVA Analysis: Player Class
+### Method under test: `addCardToHand(Card card)`
+- **TC1: add card to empty hand** ( checkmark )
```
jazlynho2028: A more specific expected output for all addCardToHand TCs would be last card in hand is _card_, instead of hand contains the specific card

docs/bva/Player.md
```diff
@@ -0,0 +1,38 @@
...
+
+- **TC3: add card to hand with five cards** ( checkmark )
+    - **State of the system**: player hand contains five cards; card is non-null
+    - **Expected output**: player hand contains six cards; hand contains specified card 
+
+- **TC4: add card to hand with duplicate cards** ( checkmark )
+    - **State of the system**: player hand has duplicate cards; card is non-null
+    - **Expected output**: player hand still has duplicate cards; hand contains specified card
+
+### Method under test: `removeCardFromHand(Card card)`
```
jazlynho2028, line 23: Add TC for remove existing card from hand with duplicate cards

jazlynho2028, line 23: Add TC for remove nonexisting card from hand with duplicate cards

docs/bva/Player.md
```diff
@@ -0,0 +1,38 @@
+# BVA Analysis: Player Class
```
jazlynho2028, line 1: Would suggest adding a "Name of the test" field under each test case to indicate which test in PlayerTests corresponds to each TC. Example:

- **TC1: player index is the same as current player index** ( :white-check-mark: )
    - **Name of the test**: onNameTag_playerStaysTheSame_noChange
    - **State of the system**: playerIndex = 0, currentPlayerIndex = 0
    - **Expected output**: no model method is called

jazlynho2028 (changes requested):
Left some comments, otherwise looks good.

docs/bva/Player.md
dongjoo224: Do you say this because the order in which the cards are added is important? Or because it's a better way to test if a card was added?

jazlynho2028: The UI reloads the whole hand every a new card is added, so you could technically add the new card wherever. But it would be more intuitive for the users to have new cards added at the end.

docs/bva/Player.md
```diff
@@ -0,0 +1,57 @@
...
```
jazlynho2028, line 1: Btw the checkmark at the end of each test case should be :checkmark:

dongjoo224: Ah got it, thanks.

docs/bva/Player.md
dongjoo224: Got it! Makes sense. I added the check in the test cases.

docs/bva/Player.md
jazlynho2028, line 1: All test cases should follow the naming convention methodUnderTest_stateOfSystem_expectedResult. For example, TC1 could be named addCardToHand_emptyHand_cardAddedToEnd

src/main/java/domain/Player.java
```diff
@@ -0,0 +1,38 @@
...
+        return this.hand.size();
+    }
+
+    public void addCardtoHand(Card card) {
+        this.hand.add(card);
+    }
+
+    public void removeCardFromHand(Card card) {
+        if (!this.hand.contains(card)) {
+            throw new IllegalArgumentException("Card not found in player hand.");
```
jazlynho2028: Strings should use the i18n .properties labels. To do this, you can replace your IllegalArgException line here with:
throw new GameException("error.cardNotInHand")

where you can implement GameException with the following definition:
public class GameException extends Exception {
    private final String key;

    public GameException(String key) {
        this.key = key;
    }

    public String getKey() {
        return code;
    }
}
Note that you wouldn't have to test GameException since it only has a default constructor and getter.

Then the test in PlayerTests would do
EasyMock.expect(player.removeCardFromHand(card))
            .andThrow(new GameException("error.cardNotInHand"));

Then later we can update the .properties files to include
error.cardNotInHand=Card not found in player hand.

And when the game class is implemented, it can handle the details of actually displaying the error.

jazlynho2028 (changes requested):

docs/bva/Player.md
dongjoo224: Got it! I fixed them in my newest commit.

src/main/java/domain/Player.java
dongjoo224: Got it. I didn't create a properties file yet, but I created the new GameException error class and used them in my test cases.

jazlynho2028 (approved):
It looks good now! Thanks for making the changes.

docs/bva/Player.md
```diff
@@ -0,0 +1,57 @@
...
+### Method under test: `addCardToHand(Card card)`
+- **TC1: add card to empty hand** ( :checkmark: )
```
jazlynho2028: small fix: use ( :white_check_mark: )

jazlynho2028: white_check_mark with colons on the side

jazlynho2028 (changes requested):

docs/bva/Player.md
dongjoo224: Got it! Fixed it in my latest commit.

jazlynho2028 (approved):
Looks good now, thanks!

src/main/java/domain/GameException.java
```diff
@@ -0,0 +1,13 @@
+package domain;
+
+public class GameException extends Exception {
```
jazlynho2028: GameException should extend RuntimeException to make it an unchecked exception

src/main/java/domain/Player.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public int getHandSize() {
+        return this.hand.size();
+    }
+
+    public void addCardtoHand(Card card) {
+        this.hand.add(card);
+    }
+
+    public void removeCardFromHand(Card card) throws GameException {
```
jazlynho2028: If GameException is an unchecked exception, you don't need to indicate "throws GameException" in method headers

src/test/java/domain/PlayerTests.java
```diff
@@ -0,0 +1,221 @@
...
+        GameException exception = assertThrows(GameException.class, () -> {
+            player.removeCardFromHand(mockNonexistingCard);
+        });
+        assertEquals("error.cardNotInHand", exception.getKey());
+
+        EasyMock.verify(mockFirstCard, mockSecondCard, mockNonexistingCard);
+    }
+
+    @Test
+    public void removeCardFromHand_multipleCardsAndMatchingCard_cardRemoved() throws GameException {
```
jazlynho2028: Same comment as above.

src/test/java/domain/PlayerTests.java
```diff
@@ -0,0 +1,221 @@
...
+        GameException exception = assertThrows(GameException.class, () -> {
+            player.removeCardFromHand(mockNonExistingCard);
+        });
+        assertEquals("error.cardNotInHand", exception.getKey());
+
+        EasyMock.verify(mockDuplicateCard, mockNonExistingCard);
+    }
+
+    @Test
+    public void removeCardFromHand_duplicateCard_oneInstanceRemoved() throws GameException {
```
jazlynho2028: Same comment as above.

jazlynho2028 (changes requested):
Sorry, I just realized GameException should extend RuntimeException to make it an unchecked exception. Everything else is good.

Luke-Meyer-hue:
Made changes to Player.java class based on more recently agreed on style of coding. Changed the GameException to IllegalStateException added a set method and adjusted some minor styling

src/main/java/domain/Player.java
```diff
@@ -1,4 +1,42 @@
...
+        hand.get(handCardIndex).toggleSelected();
+    }
+
+    public void removeCardFromHand(Card card) {
+        if (!this.hand.contains(card)) {
+            throw new IllegalStateException("error.cardNotInHand");
+        }
+        hand.remove(card);
+    }
 }
```
jazlynho2028, line 53: A public void deselectHandCards() method is needed to end a turn

src/main/java/domain/Player.java
```diff
@@ -1,4 +1,42 @@
...
+    public Player(String name) {
+        this.name = name;
+        this.hand = new ArrayList<>();
+    }
+
+    public String getName() {
+        return this.name;
+    }
+
+    public List<Card> getHand() {
```
jazlynho2028, line 20: I see in your test cases that some of them do player.getHand.getSize() and some do player.getHandSize(). If hand size is only needed in the tests, then having getHandSize() is good, but make it package private. 

Hand size hasn't been needed in the model or controller to my understanding, but if there's a place where it's needed, keep getHandSize() public.

src/test/java/domain/PlayerTests.java
```diff
@@ -0,0 +1,218 @@
...
+        IllegalStateException exception = assertThrows(IllegalStateException.class, () -> {
+            player.removeCardFromHand(mockNonExistingCard);
+        });
+        assertEquals("error.cardNotInHand", exception.getMessage());
+
+        EasyMock.verify(mockExistingCard, mockNonExistingCard);
+    }
+
+    @Test
+    public void removeCardFromHand_oneCardAndMatchingCard_handBecomesEmpty() throws GameException {
```
jazlynho2028: This method shouldn't throw an exception. Even if it does, we only throw unchecked exceptions, so no header indication is needed.

docs/bva/Player.md
```diff
@@ -0,0 +1,57 @@
...
+    - **Expected output**: player hand contains six cards; card at index 5 (last position) is the specified card instance
+
+- **TC4: add card to hand with duplicate cards** ( :white_check_mark: )
+    - **Name of the test**: addCardToHand_handHasDuplicateCards_cardAddedToEnd
+    - **State of the system**: player hand has duplicate cards; card is non-null
+    - **Expected output**: player hand still has duplicate cards; card at last index (size - 1) is the specified card instance
+
+### Method under test: `removeCardFromHand(Card card)`
+- **TC5: remove non-existing card from empty hand** ( :white_check_mark: )
+    - **Name of the test**: removeCardFromHand_emptyHand_throwsGameException
```
jazlynho2028: We're not using GameException anymore, so can you rename the test case names that still have GameException in them?

docs/bva/Player.md
```diff
@@ -0,0 +1,57 @@
...
+    - **State of the system**: player hand is empty; card to remove does not exist in hand 
+    - **Expected output**: IllegalStateException called with key "error.cardNotInHand"
+
+- **TC6: remove non-existing card from hand with one card** ( :white_check_mark: )
+    - **Name of the test**: removeCardFromHand_oneCardAndMissingCard_throwsGameException
```
jazlynho2028: See R25

docs/bva/Player.md
```diff
@@ -0,0 +1,57 @@
...
+    - **State of the system**: player hand has one card; card to remove does not exist in hand 
+    - **Expected output**: IllegalStateException called with key "error.cardNotInHand"
+
+- **TC7: remove existing card from hand with one card** ( :white_check_mark: )
+    - **Name of the test**: removeCardFromHand_oneCardAndMatchingCard_handBecomesEmpty
+    - **State of the system**: player hand has one card; card to remove does exist in hand
+    - **Expected output**: player hand does not have card anymore; player hand size is 0
+
+- **TC8: remove non-existing card from hand with more than one card** ( :white_check_mark: )
+    - **Name of the test**: removeCardFromHand_multipleCardsAndMissingCard_throwsGameException
```
jazlynho2028: See R25

docs/bva/Player.md
```diff
@@ -0,0 +1,57 @@
...
+    - **State of the system**: player hand has more than one card; card to remove does not exist in hand
+    - **Expected output**: IllegalStateException called with key "error.cardNotInHand"
+
+- **TC8: remove existing card from hand with more than one card** ( :white_check_mark: )
+    - **Name of the test**: removeCardFromHand_multipleCardsAndMatchingCard_cardRemoved
+    - **State of the system**: player hand has more than one card; card to remove does exist in hand 
+    - **Expected output**: player hand does not have card anymore; player hand size decreased by 1
+
+- **TC10: remove non-existing card from hand with duplicate cards** ( :white_check_mark: )
+    - **Name of the test**: removeCardFromHand_duplicateCardsAndMissingCard_throwsGameException
```
jazlynho2028: See R25

src/main/java/domain/GameException.java
```diff
@@ -0,0 +1,13 @@
+package domain;
```
jazlynho2028: Feel free to delete this file

src/test/java/domain/PlayerTests.java
```diff
@@ -0,0 +1,218 @@
...
+        assertEquals(finalHandSize, player.getHand().size());
+        assertEquals(mockDuplicateCard, player.getHand().get(0));
+        assertEquals(mockDuplicateCard, player.getHand().get(1));
+        assertEquals(mockNewCard, player.getHand().get(2));
+
+        EasyMock.verify(mockDuplicateCard, mockNewCard);
+    }
+
+    @Test
+    public void removeCardFromHand_emptyHand_throwsGameException() {
```
jazlynho2028: See R25

src/test/java/domain/PlayerTests.java
```diff
@@ -0,0 +1,218 @@
...
+        IllegalStateException exception = assertThrows(IllegalStateException.class, () -> {
+            player.removeCardFromHand(mockCard);
+        });
+        assertEquals("error.cardNotInHand", exception.getMessage());
+
+        EasyMock.verify(mockCard);
+    }
+
+    @Test
+    public void removeCardFromHand_oneCardAndMissingCard_throwsGameException() {
```
jazlynho2028: See R25

src/test/java/domain/PlayerTests.java
```diff
@@ -0,0 +1,218 @@
...
+        player.removeCardFromHand(mockExistingCard);
+
+        assertEquals(0, player.getHand().size());
+        assertFalse(player.getHand().contains(mockExistingCard));
+
+        EasyMock.verify(mockExistingCard);
+    }
+
+    @Test
+    public void removeCardFromHand_multipleCardsAndMissingCard_throwsGameException() {
```
jazlynho2028: See R25

src/test/java/domain/PlayerTests.java
jazlynho2028: See R25

src/test/java/domain/PlayerTests.java
```diff
@@ -0,0 +1,218 @@
...
+
+        assertEquals(maintainHandSize, player.getHand().size());
+        assertFalse(player.getHand().contains(mockCardToRemove));
+        assertSame(mockCardToKeep, player.getHand().get(0));
+
+        EasyMock.verify(mockCardToRemove, mockCardToKeep);
+    }
+
+    @Test
+    public void removeCardFromHand_duplicateCardsAndMissingCard_throwsGameException() {
```
jazlynho2028: See R25

src/test/java/domain/PlayerTests.java
jazlynho2028: See R128

jazlynho2028 (changes requested):
Very nice. Just some minor details with missed naming adjustments and error headers. Also would like to add a public void deselectHandCards method.

src/main/java/domain/Player.java
```diff
@@ -1,4 +1,42 @@
...
+
+    public int getHandSize() {
+        return this.hand.size();
+    }
+
+    public void addCardToHand(Card card) {
+        hand.add(card);
+    }
+
+    public void setIsSelectedOfHandCardToOpposite(int handCardIndex) {
```
jazlynho2028: A better name would be toggleSelectedHandCardAt(int index)

jazlynho2028: should throw an exception if handCardIndex is out of bounds

src/main/java/domain/Player.java
Luke-Meyer-hue: I'll keep getHandSize because I think its important to the tests, at least from what I saw initially.

src/test/java/domain/PlayerTests.java
Luke-Meyer-hue: Sorry, I didn't see that while adjusting the other removeCardFromHand tests that have to do with the illegalStateException. I'll remove it from the files.

src/main/java/domain/Player.java
Luke-Meyer-hue: Added public void deselectHandCards() with a BVA and test cases in order to deselect all cards in the hand before moving to next turn

dongjoo224 (commented):
Maybe I'm not seeing the change, but GameException class can be deleted like Jazlyn said. Once that's removed everything looks good.

src/main/resources/languages/strings_Spanish.properties
```diff
@@ -9,6 +9,7 @@ error.drawFromPile=No se pudo robar del mazo.
 error.startGame=No se pudo iniciar el juego.
 error.emptyPlayerList=La lista de jugadores no puede estar vac�a.
 error.negativeDrawCount=El contador de robos no puede ser menor que cero.
+error.cardNotInHand=No se puede quitar una carta que no est� en la mano.
 
 startScreen.English=INGLÉS
```
jazlynho2028, line 14: Would you mind making a PR to resolve the weird characters appearing for startScreen.English and startScreen.Spanish in the spanish.properties file?

src/main/java/domain/Player.java
Luke-Meyer-hue: Adjusted method to have name toggleSelectedHandCardAt and throw an exception. Created BVA and test cases for this method.

jazlynho2028 (approved):
Looks good now, thanks!

PR #35, opened 2026-05-04
Game Setup Phase: TurnManager class

### LINKED ISSUE #15: Game setup phase: TurnManager class

### LINKED ISSUE #83: One turn of the game: TurnManager class
Fields

int currentPlayerIndex
List<Player> players
int roundCounter
int currentDrawCount

Methods

public TurnManager(final List<Player> players)
public int getCurrentPlayerIndex()
public Player getCurrentPlayer()
public int getCurrentDrawCount()
public int getStartingPlayerIndex()
public int getRoundCount()
public void setCurrentPlayerIndex(int newPlayerIndex)
public void incrementDrawDraw()
public void decrementDrawCount()
public void incrementRound()
public void advanceTurn()

dongjoo224:


docs/bva/TurnManager.md
```diff
@@ -0,0 +1,11 @@
+# BVA Analysis: TurnManager Class
+### Method under test: `getCurrentPlayerIndex()`
+- **TC1: first player's index before any turns taken with list of two players** ( checkmark )
```
jazlynho2028: ( checkmark ) should be ( ✅ ) to display the icon as expected in .md format 

white_check_mark with colons on the sides

src/main/java/domain/TurnManager.java
```diff
@@ -0,0 +1,46 @@
...
+public final class TurnManager {
+    //    private int startingPlayerIndex;
+    private int currentPlayerIndex;
+    private final List<Player> players;
+    //    private int roundCounter;
+    //    private int currentDrawCount;
+
+    public TurnManager(final List<Player> players) {
+        if (players == null) {
+            throw new IllegalArgumentException("Player list cannot be null.");
```
jazlynho2028: Same comment as the Player class about using .properties and GameException for error strings.  

error.nullPlayers=Player list cannot be null.

jazlynho2028 (changes requested):
Just left comments for some minor changes. Looks good otherwise!

src/main/java/domain/TurnManager.java
```diff
@@ -0,0 +1,46 @@
+package domain;
+
+import java.util.ArrayList;
```
jazlynho2028: Remove this import if it's not used

docs/bva/TurnManager.md
dongjoo224: Got it!

src/main/java/domain/TurnManager.java
dongjoo224: Must have forgotten to remove it as I was editing the file. Thanks!

src/main/java/domain/TurnManager.java
dongjoo224: Understood. Made the changes in my most recent commit.

jazlynho2028 (changes requested):
If you change these two lines in build.gradle.kts

pitest {
    targetClasses.set(listOf("domain.*", "datasource.*"))
    targetTests = setOf("Code*")

to

pitest {
    targetClasses.set(listOf("domain.TurnManager"))
    targetTests = setOf("domain.TurnManagerTests")

Then run the gradle "test" task and "jacocoTestReport", you can see that one of the elements have 50% branch coverage. 

You probably implemented enough just to finish GameSetUp, so you can just worry about the n/a coverage for empty methods the coming week, but you should have a constructor test that covers the null players possibility since you have a check for that.

docs/bva/TurnManager.md
```diff
@@ -0,0 +1,71 @@
...
+  - **State of the system**: currentDrawCount = 0
+  - **Expected output**: getCurrentDrawCount() = -1
+
+- **TC7: Draw count incremented on nonzero draw count** ( :white_check_mark: )
+  - **Name of the test**: decrementDrawCount_fromNonZero_success
+  - **State of the system**: currentDrawCount = -1
+  - **Expected output**: getCurrentDrawCount() = -2
+
+### Method under test: `incrementRound()`
+- **TC8: Round incremented from its minimum baseline** ( :white_check_mark: )
```
jazlynho2028, line 41: Can you throw an IllegalStateException to disallow negative values for drawCount?

docs/bva/TurnManager.md
```diff
@@ -0,0 +1,71 @@
...
+  - **State of the system**: roundCounter = 0
+  - **Expected output**: getRoundCounter() = 1
+
+- **TC9: Draw count incremented on nonzero draw count** ( :white_check_mark: )
+  - **Name of the test**: incrementRound_fromNonZero_success
+  - **State of the system**: roundCounter = 1
+  - **Expected output**: getRoundCounter() = 2
+
+### Method under test: `advanceTurn()`
+- **TC10: Advance turn away from starting baseline** ( :white_check_mark: )
```
jazlynho2028: No change in implementation or tests needed, but keep in mind that when Game does startGame(), it will increment the roundCounter so that rounds are indexed starting with 1.

src/main/java/domain/TurnManager.java
```diff
@@ -0,0 +1,70 @@
...
+    private static final int STARTING_PLAYER_INDEX = 0;
+
+    private int currentPlayerIndex;
+    private final List<Player> players;
+    private int roundCounter;
+    private int currentDrawCount;
+
+    public TurnManager(final List<Player> players) {
+        if (players.isEmpty()) {
+            throw new IllegalArgumentException("Players list cannot be empty");
```
jazlynho2028: replace with

throw new IllegalArgumentException("error.emptyPlayerList")

then add the key "error.emptyPlayerList" and message "Players list cannot be empty." to both english and spanish .properties files.

Then when you test, treat the expectedMsg string to be "error.emptyPlayerList" instead of the actual message

src/test/java/domain/TurnManagerTests.java
```diff
@@ -0,0 +1,305 @@
...
+        List<Player> emptyPlayersList = new ArrayList<>();
+
+        IllegalArgumentException exception = assertThrows(
+                IllegalArgumentException.class,
+                () -> {
+                    new TurnManager(emptyPlayersList);
+                }
+        );
+
+        assertEquals("Players list cannot be empty", exception.getMessage());
```
jazlynho2028: See above comment

docs/bva/TurnManager.md
```diff
@@ -0,0 +1,71 @@
...
+  - **Name of the test**: advanceTurn_fromIndexZero_currentPlayerIndexBecomesOne
```
jazlynho2028: Might be good to test advance turn with min and max number of players rather than 3 players. You can use parameterized testing like in lab 4 DiceTests to combine these into one test.

jazlynho2028 (changes requested):

docs/bva/TurnManager.md
Luke-Meyer-hue: Yes, IllegalStateException gets thrown when decrementDrawCount gets called while drawCount is <= 0 and throughs an exception. The exception has been added to the property files

src/main/java/domain/TurnManager.java
Luke-Meyer-hue: The error has been replaced and the message has been put into both the property files in the correct languages

docs/bva/TurnManager.md
Luke-Meyer-hue: Used parameterized testing to test with 2 to 4 players in each of the advance turns

docs/bva/TurnManager.md
```diff
@@ -0,0 +1,71 @@
...
+- **TC5: Draw count incremented on positive draw count** ( :white_check_mark: )
+  - **Name of the test**: incrementDrawCount_fromOne_success
+  - **State of the system**: currentDrawCount = 1
+  - **Expected output**: getCurrentDrawCount() = 2
+
+### Method under test: `decrementDrawCount()`
+- **TC6: Decrementing from zero throws state exception*** ( :white_check_mark: )
+  - **Name of the test**: decrementDrawCount_fromZero_throwsIllegalStateException
+  - **State of the system**: currentDrawCount = 0
+  - **Expected output**: getCurrentDrawCount() = -1
```
jazlynho2028: Expected output should be IllegalStateException "error.negativeDrawCount"

src/test/java/domain/TurnManagerTests.java
```diff
@@ -0,0 +1,298 @@
...
+        turnManager.incrementRound();
+
+        int actual = turnManager.getRoundCounter();
+
+        assertEquals(initialRoundCounter + 2, actual);
+
+        EasyMock.verify(mockPlayer1);
+    }
+
+    @ParameterizedTest
```
jazlynho2028, line 210: You could further parameterize all the advanceTurn tests into one

jazlynho2028 (changes requested):
Thanks for making the changes! Just commented on two small things

jazlynho2028 (approved):
Looks good now!

PR #36, opened 2026-05-04
Game Setup Phase: Game class

### LINKED ISSUE #17: Game setup phase: Game class

dongjoo224:


docs/bva/Game.md
```diff
@@ -0,0 +1,65 @@
+# BVA Analysis: Game Class
+### Method under test: `startGame()`
```
jazlynho2028, line 78: I see that your Game.java constructor checks for valid playerNames size, so you should write test cases for the constructor (since it's not a default constructor) so you don't need TC4-5 for startGame().

Then Game constructor should: check if playerNames is valid, initialize players, isFaceUp=False, canDraw=False, drawPile, discardPile, and isGameOngoing=False or GamePhase (we probably don't need GamePhase at this point if we have isGameOngoing. I know I put it in the design but it might be overcomplicated to have a GamePhase)

Then startGame() initializes TurnManager and sets the appropriate booleans.

docs/bva/Game.md
```diff
@@ -0,0 +1,65 @@
...
+
+- **TC4: start game with too little players** ( x )
+  - **State of the system**: Game constructed with 1 player name; isGameOngoing is false
+  - **Expected output**: IllegalArgumentException
+
+- **TC5: start game with too many players** ( x )
+  - **State of the system**: Game constructed with 5 player names; isGameOngoing is false
+  - **Expected output**: IllegalArgumentException
+
+### Method under test: `getIsBeforeDraw()`
```
jazlynho2028: I'm guessing you'll remove this method since we can just change the boolean canDraw

docs/bva/Game.md
jazlynho2028: Let me know if you want help with doing BVA or implementation for Game or TurnManager

docs/bva/Game.md
dongjoo224: Yep. I'll just remove it now to keep it simple.

docs/bva/Game.md
dongjoo224: Thanks for the suggestions! I revised the BVA analysis in my latest commit incorporating these changes.

jazlynho2028 (changes requested):
Would you mind adding a List<String> cardIds parameter to Game so that it can pass it to DeckBuilder? This is so that DeckBuilder can assign cardIds to each card.

dongjoo224:
I keep running into errors with returning references to mutable objects in my getters. All other errors are resolved but these, specifically with the getDiscardPile() and getDrawPile() methods. Other than that, this PR is ready for review.

src/main/java/domain/Game.java
```diff
@@ -0,0 +1,108 @@
+package domain;
```
jazlynho2028: Please merge main into your branch asap so that you can run jacoco and pitest on your test cases. Right now, your tests for Game have 100% branch coverage but do not kill all mutants. To isolate a class for pitest, temporarily change the following two lines in build.gradle.kts:

pitest {
    targetClasses.set(listOf("domain.*", "ui.*Controller"))
    targetTests.set(listOf("domain.*", "ui.*ControllerTests"))

to:

pitest {
    targetClasses.set(listOf("domain.Game"))
    targetTests.set(listOf("domain.GameTests"))

src/main/java/domain/TurnManager.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public final class TurnManager {
+    private final List<Player> players;
+    private int currentPlayerIndex;
+
+    public TurnManager(List<Player> players) {
+        if (players == null || players.isEmpty()) {
+            throw new IllegalArgumentException("Player collection cannot be null or empty.");
```
jazlynho2028: Is there a reason why the exception here wasn't handled with a GameException?

PR #40, opened 2026-05-04
Week 5 Report

Sunw00ha:
updated week 5 report


jazlynho2028:
Since shuffling is random, it is not guaranteed that the ordering would be changed. Maybe take a look at the example exploding kittens command line project to see how they did testing for that. Also do you mind committing the BVA in the corresponding wip branch next time separate from the report next time?

PR #53, opened 2026-05-10
Wip card metadata

### LINKED ISSUE #50: Card Metadata
Every card in Exploding Kittens has a unique combination of title, subtitle, description, and image. This metadata will be stored in a .json file in src/main/java/resources. A corresponding CardMetadata class and CardMetadataLoader will need to be implemented to load the .json file into a usable format to be displayed in the UI.

jazlynho2028:
Every card in Exploding Kittens has a unique combination of title, subtitle, description, and image. This pr includes a CardMetadata class (with title, subtitle, description, imageUrl) that the UI can use to display. A HashMap that maps a card's id to its corresponding CardMetadata instance is added in AssetManager, and is populated (and can be accessed just like the images and svgPath hashmaps) when all other fonts/images/stylesheet/icons are loaded.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -19,8 +19,10 @@ repositories {
 }
 
 dependencies {
+    // checkstyle
     implementation("com.puppycrawl.tools:checkstyle:10.18.2")
 
+    // test
     testImplementation(platform("org.junit:junit-bom:5.10.0"))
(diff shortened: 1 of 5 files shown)


dongjoo224 (approved):
Everything looks good.

PR #64, opened 2026-05-15
Replaced three card types in .properties files

jazlynho2028:
Replaced NOPE, RAISING HECK, and CLEAN UP with FERAL CAT, SUPER SKIP, SHUFFLE.


Luke-Meyer-hue (approved):
Everything looks good

PR #70, opened 2026-05-23
Game Setup Phase: DeckBuilder class

### LINKED ISSUE #12: Game setup phase: Deckbuilder class
setup:
- deck before distributing cards to players
- method to shuffle deck
- deck after distributing cards to players (deck - top (n*5) cards)

### LINKED ISSUE #84: One turn of the game: DeckBuilder class
Fields

private static List<Card> cards

Methods

public static Deck buildDeckWithoutExplodeAndAddDefuse(int numPlayers)
static Deck initializeFullDeck()
public static String createCardId(CardType type, int num)
static int calculateDefusesToAdd(int numPlayers)

ejeankuo:


docs/bva/DeckBuilder.md
```diff
@@ -0,0 +1,56 @@
+# BVA Analysis: DeckBuilder Class
```
jazlynho2028, line 1: See Player class for an example how to handle exceptions in classes and tests so that they display accordingly using the i18n implementation.

docs/bva/DeckBuilder.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Method under test: `buildDeckWithoutExplodeAndAddDefuse(int numPlayers)`
+- **TC1: Less than minimum allowed players ** ( :white_check_mark: )
```
jazlynho2028: If you have time to change this, Game already handles players being 2-4 since this is a Game rule. The only exceptions DeckBuilder would handle would be things that only apply to it, like if the number of players is < 0 or > 5 so that there is no negative result in any computation in the class.

docs/bva/DeckBuilder.md
jazlynho2028, line 1: TC should be numbered starting at 1 for each class, not for each method. I would also suggest adding an extra line under each test to indicate the corresponding test function.

Ex. 
### Method under test: `StartController(StartView view)`
- **TC1: constructor is called ** ( :white_check_mark: )
  - **Name of the test**: constructor_called_success
  - **State of the system**: N/A
  - **Expected output**: called view.bindUI()

src/main/java/domain/Card.java
```diff
@@ -0,0 +1,14 @@
+package domain;
```
jazlynho2028, line 1: Luke and i just realized that the Card object needs to take in both a string ID and the CardType enum. Would you be okay with changing DeckBuilder to take in a list of card id strings and use that to pass in both the id and CardType to the Card constructor? Or is there another way to do this that you think would be better? The app already stores a map of card ids to metadata, so this way it can just pass a list of ids based on that map to Game during initialization

jazlynho2028 (changes requested):
Just a couple changes about error handling and implementation. Looks good otherwise.

jazlynho2028 (changes requested):
Please pull from main to get the updated build.gradle.kts file so you can check your test coverage and mutants in pitest. To isolate specific classes, change the following two lines in build.gradle.kts:

pitest {
    targetClasses.set(listOf("domain.*", "ui.*Controller"))
    targetTests.set(listOf("domain.*", "ui.*ControllerTests"))

to:

pitest {
    targetClasses.set(listOf("domain.DeckBuilder"))
    targetTests.set(listOf("domain.DeckBuilderTests"))

src/main/java/domain/Card.java
ejeankuo: Ah i see, yes got it, will update deckbuilder class to also take in string IDs

docs/bva/DeckBuilder.md
ejeankuo: Got it, ill change then to only throw an exception if the number of players produces a negative number of defuses.

src/test/java/domain/DeckBuilderTests.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import static domain.GameConstants.*;
+import static org.junit.jupiter.api.Assertions.*;
+
+public class DeckBuilderTests {
+
+    @Test
+    void buildDeck_MinAllowedPlayers_ReturnsCorrect59CardDeck() {
```
jazlynho2028: You could parameterize buildDeck_minAllowedPlayers_returnsCorrect59CardDeck and buildDeck_maxAllowedPlayers_returnsCorrect57CardDeck into one test with parameters numPlayers and expectedDeckSize

src/main/java/domain/DeckBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    private static void addCards(List<Card> destinationList, CardType type, int numToAdd) {
+        for (int i = 1; i <= numToAdd; i++) {
+            String cardId = createCardId(type, i);
+            destinationList.add(new Card(cardId, type));
+        }
+    }
+
+    public static String createCardId(CardType type, int num) {
```
jazlynho2028: You should do BVA and write tests for this createCardId method.

src/main/java/domain/DeckBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+        String cardTypeNameWithoutUnderscore = type.name().replace("_", "");
+
+        return String.format(
+                "%s_%d",
+                cardTypeNameWithoutUnderscore,
+                num
+        );
+    }
+
+    static int calculateDefusesToAdd(int numPlayers) {
```
jazlynho2028, line 65: The number of defuses in the game always stays the same, unlike exploding kittens, so just NUM_DEFUSES would suffice for the name of MAX_DEFUSES_COUNT

src/test/java/domain/DeckBuilderTests.java
```diff
@@ -0,0 +1,77 @@
+package domain;
```
jazlynho2028, line 1: The tests you wrote are integration tests, not unit tests, because they are mocking all the non DeckBuilder objects. Your tests should all pass even if all Deck and Card methods are empty/return static values.

src/main/java/domain/DeckBuilder.java
jazlynho2028: Follow TDD and write BVA before writing this method directly.

src/test/java/domain/DeckBuilderTests.java
jazlynho2028: Instead of checking the deck size, it is more important that we have the expected numbers of each card type. So you should define the following constants in GameConstants and use them in your DeckBuilder and tests

NUM_DEFUSE = 5
NUM_EXPLODING_KITTEN = 3
NUM_CAT_CARD = 4
NUM_ATTACK = 3
etc...

docs/bva/DeckBuilder.md
```diff
@@ -0,0 +1,29 @@
+# BVA Analysis: DeckBuilder Class
+
```
dongjoo224, line 2: Following the format of the other test files, there should be a "Name of the test" bullet under each test as well. 

For instance for TC1: 
- **TC1: Minimum allowed players ** ( :white_check_mark: )
    - **Name of the test**:  buildDeck_MinAllowedPlayers_ReturnsCorrect59CardDeck
    - **State of the system**: numPlayers = 2
    - **Expected output**: deck has 59 cards, of which 3 are "DEFUSE"'s. deck is shuffled.

src/main/java/domain/DeckBuilder.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.Deque;
+import java.util.List;
+
+import static domain.GameConstants.*;
+
+public class DeckBuilder {
+
+
+
+    public static Deck buildDeckWithoutExplodeAndAddDefuse(int numPlayers) {
```
jazlynho2028: Since this method uses external classes like Card and Deck, you should make it non static and let the main app create a DeckBuilder object to make it testable. Then DeckBuilder stores a Deck object and populates it in the constructor. Then you can test the DeckBuilder constructor for behavior.

src/main/java/domain/DeckBuilder.java
jazlynho2028: This method is fine to be static because it does not use any external classes

jazlynho2028 (changes requested):
Needs some major changes to make DeckBuilder testable

src/main/java/domain/DeckBuilder.java
ejeankuo: I can definitely see your point, but I was also thinking that just saying NUM_DEFUSES might be confusing because its not clear if its the number of defuses currently in the draw pile, how many defuses to add to the draw pile based on number of players, etc, so I thought naming it MAX_DEFUSES_COUNT or TOTAL_DEFUSES_IN_GAME would be more clear. I'm open to your thoughts if you think otherwise!!

src/test/java/domain/DeckBuilderTests.java
ejeankuo: Great point, thanks for the idea. Will fix to check for expected number.

src/main/java/domain/DeckBuilder.java
jazlynho2028: Good point. I like NUM_DEFUSES_IN_GAME because you can use the same name formatting for all the other card types, so you can define NUM_ATTACKS_IN_GAME, etc.

src/test/java/domain/DeckBuilderTests.java
ejeankuo: I ran into some complications trying to mock Deck and Cards for DeckBuilder, and I think making DeckBuilder use sociable unit tests might be a better idea? Because DeckBuilder is a factory for instantiating/populating a Deck object, I think that mocking Deck could potentially defeat the purpose of the tests to verify if the builder is successfully creating a deck (bc mocking can only verify that a method was called, but we need to verify the actual card contents of the final product?). But I'm also conflicted because I can see your point that if Deck/Card fails, then DeckBuilder will also fail. Unless you feel strongly about this, maybe I can ask Prof. Yiji what she thinks?

ejeankuo: Update: refactored the code to be able to write unit tests for helper functions using partial mocking

src/test/java/domain/DeckBuilderTests.java
```diff
@@ -0,0 +1,175 @@
...
+        verifyCardTypeGroup(baseCards, CardType.CAT_CARD_1, NUM_CAT_CARD_IN_GAME, "CATCARD1");
+        verifyCardTypeGroup(baseCards, CardType.CAT_CARD_2, NUM_CAT_CARD_IN_GAME, "CATCARD2");
+        verifyCardTypeGroup(baseCards, CardType.CAT_CARD_3, NUM_CAT_CARD_IN_GAME, "CATCARD3");
+        verifyCardTypeGroup(baseCards, CardType.CAT_CARD_4, NUM_CAT_CARD_IN_GAME, "CATCARD4");
+    }
+
+    @Test
+    void calculateDefusesToAdd_NegativeDefuses_ThrowsException() {
+        assertThrows(IllegalArgumentException.class, () -> {
+            DeckBuilder.calculateDefusesToAdd(INVALID_PLAYER_COUNT);
```
jazlynho2028: INVALID_PLAYER_COUNT is not explicit because there are many invalid player counts, not just 6. The test would be easier to understand if you passed in NUM_DEFUSES_IN_GAME + 1 instead.

src/test/java/domain/DeckBuilderTests.java
```diff
@@ -0,0 +1,175 @@
...
+        });
+    }
+
+    @Test
+    void calculateDefusesToAdd_MinimumPlayers_ReturnsThree() {
```
jazlynho2028, line 140: Instead of doing EXPECTED_DEFUSE_COUNT_2_PLAYERS and EXPECTED_DEFUSE_COUNT_4_PLAYERS, you should do NUM_DEFUSES_IN_GAME - MIN_PLAYERS and NUM_DEFUSES_IN_GAME - MAX_PLAYERS for a more explicit definition that follows the game rule instructions. You can also parameterize the two tests with a numPlayers argument if you'd like.

src/main/java/domain/GameConstants.java
```diff
@@ -10,12 +10,37 @@ private GameConstants() { }
...
+	public static final int NUM_SKIP_IN_GAME = 3;
+	public static final int NUM_CLONE_IN_GAME = 3;
+	public static final int NUM_SWAP_TOP_AND_BOTTOM_IN_GAME = 3;
+	public static final int NUM_DRAW_FROM_THE_BOTTOM_IN_GAME = 3;
+	public static final int NUM_FERAL_CAT_IN_GAME = 4;
+	public static final int NUM_SEE_THE_FUTURE_IN_GAME = 4;
+	public static final int NUM_SHUFFLE_IN_GAME = 4;
+	public static final int NUM_TARGETED_ATTACK_IN_GAME = 4;
+	public static final int NUM_CAT_CARD_IN_GAME = 4;
+
```
jazlynho2028: In my opinion, these 6 constants shouldn't be needed. If a new card is added to the game, you'd have to add both a NUM_NEW_CARD_IN_GAME constant and change the EXPECTED_BASE_DECK_SIZE, EXPECTED_DECK_SIZE_4_PLAYERS, and EXPECTED_DECK_SIZE_2_PLAYERS constants. It is enough in the tests to verify that a correct number of each CardType is added to the game. Note that there is no risk of adding additional unexpected cards because CardType is an enum that already specifies all possible cards.

jazlynho2028: See further explanation in DeckBuilderTests comments

jazlynho2028 (changes requested):
Thanks for making all the changes! Just a couple things about constants now.

src/test/java/domain/DeckBuilderTests.java
ejeankuo: Great point, thanks for pointing that out

src/test/java/domain/DeckBuilderTests.java
ejeankuo: Ah great point, it would definitely be helpful to be able to see how the expected defuse count was calculated using player count. Will alter!

src/main/java/domain/GameConstants.java
ejeankuo: I can see your point about difficulty in maintenance if we were to add new cards, but I also wonder if in the case of action cards where we need to randomly insert into the deck using an index, that having the size of the deck validated first could be useful. I'm thinking I can get rid of the constants EXPECTED_BASE_DECK_SIZE etc. and instead verify the size of the deck by summing up NUM_ATTACKS + NUM_SHUFFLES, etc... rather than having a predefined EXPECTED_BASE_DECK_SIZE constant, like you suggested.

jazlynho2028 (approved):
Looks good now!

PR #77, opened 2026-05-29
Wip game revision

### LINKED ISSUE #17: Game setup phase: Game class

### LINKED ISSUE #30: One turn of the game: Requirements
**User Story:**

As a player of Exploding Kittens, I want to be able to see my cards clearly, know when I can play cards or need to draw cards, and have effects applied automatically, so that I don't accidentally make a wrong move in the game. 

**Acceptance Criteria**

- Player 1 must have either drawn a card or played a card that allowed them to not draw in this turn before they can end their turn.
- All played cards must be added into the discard pile.
- All drawn cards must be removed from the deck pile.
- The card drawn by Player 1 must be from the top of the initialized deck (or bottom depending on the play)
- Player 1 must not be allowed to play any more cards after drawing from the deck.
- Only valid and non-empty effect cards may be played
   - Overlapping effects played one at a time (ex. playing a "Skip" then an "Attack") is allowed. In this example, the overall effect would just be equivalent to an "Attack".

**Use Case 1: Play one turn of the game**

Actor: Player

Preconditions: The game application is launched. Deck and player hands are initialized. The game is ready to play.

Main Flow:

1. Player 1 clicks "Reveal" to view their hand.
- Note that player tabs now indicate the current player's turn. They are fixed and cannot be clicked.
2. Player 1 selects the card(s) they want to play. 
- multi-select is only allowed for cards of the same type for two-of-a-kind or three-of-a-kind plays).
3. Player 1 clicks "Play Card(s)" to play their selected card(s).
- If invalid card(s) are selected (ex. 1 cat card; 1 defuse card (before exploding kitten is drawn); 2 different type cards), the button will stay disabled.
4. Played card(s) are added to the discard pile.
- If multiple cards are played at once, only one of them will be visible in the discard pile UI.
5. Effect of the played card(s) activates.
- Specific details depends on the card(s) played.
6. Player 1 repeats steps 2-5 until they no longer want to play cards for the current turn.
7. Player 1 clicks the draw pile.
8. The top card from the draw pile is added to Player 1's hand.
- "Play Card(s)" button disables.
- "End Turn" button enables.
9. Player 1 clicks "End Turn" to end their turn.

Alternate Flows:
2.a. Player 1 chooses not to play any cards this round.
2.a1. Resumes at step 7.

7.a. "End Turn" button enables because Player 1 played one or more of the following: 
- "Attack"
- "Skip"
- "Raising Heck" and kept the card.
- "Targeted Attack"
- "Godcat" or "Clone" as any of the above cards.
7.a1. Resumes at step 9.

8.a. Player 1 played "Draw from the Bottom".
8.a1. The bottom card from the draw pile is added to Player 1's hand.
- "Play Card(s)" button disables.
- "End Turn" button enables.
8.a2. Resumes at step 9.

9.a. Player 1 drew an "Exploding Kitten".
- "End Turn" button stays disabled.
9.a1. Player 1 plays "Defuse".
- "Play Card(s)" button disables.
- "End Turn" button enables.
9.a2. Resumes at step 9.

**Postconditions:**

- Effects of card(s) played are applied to applicable players.
- Player 1 has a new card or played a card that allowed them to not draw in this turn.
- The game is ready for the second turn.

**Design/Flow**

Image

Image

Image

Image

Image

Image

Image

### LINKED ISSUE #82: One turn of the game: Game class
Fields:

List<Player> players
Deck drawPile
Deck discardPile
TurnManager turnManager

boolean isGameOngoing
boolean isFaceUp

Methods:

public void Game(List<Player> players, Deck drawPile, Deck discardPile, TurnManager turnManager)
public void startGame()
public List<String> getPlayerNames()
public int getCurrentPlayerIndex()
public int getStartingPlayerIndex()
public List<String> getCurrentPlayerHandIds()
private Player getCurrentPlayer()
public boolean canPlaySelected()
public boolean canEndTurn()
public boolean isDrawPileEmpty()
public boolean getIsGameOngoing()
public boolean getCanDraw()
public void changeCurrentPlayerIndex(int newPlayerIndex)
public void setFaceUpToFalse()
public void drawFromPile()
public void toggleFaceUp()
public void toggleSelectedPlayerCardAt(int handCardIndex)
public void advanceTurn()

jazlynho2028:
Reconducted BVA analysis for Game class. Includes following refactors:
- Inject List<Player> players, Deck drawPile, Deck discardPile, TurnManager turnManager into Game.java constructor and use suppressions to resolve storing mutable object and throwing in constructor errors
- Removed GameData class
- Added advanceTurn method for one turn of game phase


src/main/java/domain/Game.java
```diff
@@ -1,137 +1,192 @@
...
+    private List<Player> players;
+    private Deck drawPile;
+
+    private boolean isGameOngoing;
+    private boolean isFaceUp;
+
+    private int roundCount;
+    private int drawCount;
+
+    private static final List<CardType> UNPLAYABLE_TYPES = List.of(
```
jazlynho2028, line 22: Note that Cat cards are playable, it's just that they are only playable when two or three are played at once. Also I would recommend treating CardType.FERAL_CAT as a cat card too. Although I guess that would mean you're helping to implement the cat cards and feral cat instead of Ejean at this point. You can discuss with her and see if you want to exchange some cards with her to implement.

src/main/java/domain/Game.java
```diff
@@ -1,137 +1,192 @@
...
-        list.add("TARGETEDATTACK_2");
-        list.add("TARGETEDATTACK_3");
-        list.add("TARGETEDATTACK_4");
-        list.add("WINNERWINNERCATNIPDINNER_1");
-        list.add("RAGEBAIT_1");
-        list.add("RECYCLE_1");
-        list.add("DOUBLEUP_1");
-        list.add("MILDDRAW_1");
-        return list;
+        return players.get(turnManager.getCurrentPlayerIndex()).getHandIds();
```
jazlynho2028: Would recommend adding a private Player getCurrentPlayer that calls players.getCurrentPlayerIndex()

Also note that you have a getCurrentPlayerIndex() method already, so you don't have to do turnManager.getCurrentPlayerIndex() anymore

jazlynho2028: Actually scratch that. You can just use the TurnManager method and do turnManager.getCurrentPlayer()

src/main/java/domain/Game.java
jazlynho2028: Resolved over messages: canPlaySelected should just only return true for selectedCards of size one for now. Return false for exploding kitten, defuse, cat card, or feral cat. Return true for all other types

dongjoo224: Sounds good. I edited the BVA and modified the implementation to reflect this, focusing on valid single cards. The method treats FERAL_CAT as an unplayable card for now, and we can change it after we implement the card's functionality.

src/main/java/domain/Game.java
dongjoo224: Good catch! I decided to change it where getCurrentPlayerHandIds() now calls the existing getCurrentPlayer() method which now calls the existing getCurrentPlayerIndex() method, to get rid of repetition and to utilize the methods we already have. And since we already have a getCurrentPlayer() for Game I just used that.

jazlynho2028: If Game's getCurrentPlayer calls turnManager,getCurrentPlayer(), Game's getCurrentPlayer()  only needs one test case that verifies turnManager.getCurrentPlayer() is called.

jazlynho2028: A good rule of thumb is: if a Game method calls a method that belongs to another class, for example, player.toggleSelected(), the game test for that method only needs to verify that player.toggleSelected() is called, and it can't and shouldn't verify that player.getHand().get(0).getIsSelected() is true or false. Because the Player class's tests already verify that state whenever player.toggleSelected() is called.

dongjoo224: Got it that makes complete sense. I'll reduce the getCurrentPlayer() method's BVA analysis to one test case that accounts for this.

dongjoo224:
There's two spotbugs errors from other classes, not Game. And there are checkstyle errors in the GameTests.java file due to some lines being longer than the 100 character limit, but I'm unsure of how to shorten those specific lines.

jazlynho2028:
> There's two spotbugs errors from other classes, not Game. And there are checkstyle errors in the GameTests.java file due to some lines being longer than the 100 character limit, but I'm unsure of how to shorten those specific lines.

Resolved spotbugs and checkstyle errors

Luke-Meyer-hue (approved):
Everything seems ok to me, the code all looks good

PR #91, opened 2026-05-30
Wip player revisions

dongjoo224:


Luke-Meyer-hue:
The BVA looks good, the tests seem fine. The code looks good as well. The only thing I would say is that there are actual card Ids in resource/card-metadata.json, the tests work regardless so I'm not sure if it matters

src/test/java/domain/PlayerTests.java
```diff
@@ -340,4 +342,190 @@ public void toggleSelectedHandCardAt_indexGreaterThanHandSize_callsException() {
...
+        Player player = new Player("Alice");
+
+        List<String> handIds = player.getHandIds();
+
+        assertEquals(0, handIds.size());
+    }
+
+    @Test
+    public void getHandIds_oneCard_returnsListWithOneId() {
+        final String cardId = "card-1";
```
Luke-Meyer-hue: I'm not sure if this would be necessary to change as the tests work without it, but you could use some of the actual cardIds from resource/card-metadata.json

Luke-Meyer-hue (approved):

src/test/java/domain/PlayerTests.java
dongjoo224: Got it. Good point, it's probably better to be as specific as possible. Made the changes in my latest commit.

docs/bva/Player.md
```diff
@@ -111,4 +111,51 @@
...
+  - **Name of the test**: getSelectedCards_multipleCardsSomeSelected_returnsOnlySelectedCards
+  - **State of the system**: player hand has three cards; some but not all have isSelected = true (two cards have isSelected = true, one card has isSelected = false)
+  - **Expected output**: returns a list containing only the two selected cards; the one unselected card is not included
+
+- **TC25: Get selected cards when more than one card is selected** ( :white_check_mark: )
+  - **Name of the test**: getSelectedCards_multipleCardsAllSelected_returnsAllCards
+  - **State of the system**: player hand has three cards; every card has isSelected = true
+  - **Expected output**: returns a list containing all three cards in the hand
+
+### Method under test: `getHandIds()`
```
jazlynho2028, line 147: Could add one more TC for the case if there's duplicate ids. Expected behavior is to have those duplicates in the returned list

jazlynho2028 (commented):
Could add one more TC. Otherwise looks good!

docs/bva/Player.md
dongjoo224: Yes, got it! Just added it in my latest commits.

PR #121, opened 2026-06-02
Card Implementation: Catomic Bomb

### LINKED ISSUE #109: Game Card-CatomicBomb
When played, this card goes to the discard pile and all Exploding Kittens in the deck go to the top of the deck. After this card is played, the current turn is immediately ended and the next player starts.

dongjoo224:
The Catomic Bomb card when played puts all exploding kittens in deck currently on top and immediately advances to the next player.


src/main/java/domain/Game.java
```diff
@@ -298,7 +299,26 @@ void applySeeTheFuture() {
     }
 
     void applyCatomicBomb() {
```
jazlynho2028, line 368: I just left a review for someone's card to add an addCardToBottom method to Deck, so you can use that when they're done implementing it.

jazlynho2028 (changes requested):
Looks good! See comment about new Deck method.

src/main/java/domain/Game.java
jazlynho2028: you can add an empty method for it too if you want to merge this sooner.

dongjoo224: That was me haha. I made the changes to the Swap Top and Bottom card adding the addCardToBottom(), so I'll wait for that PR to be merged into main before I change this one. These changes should be quick anyway.

jazlynho2028 (approved):
Looks good! Would it be okay to merge my wip-godcat-revisions into main first so that the conflicts are easier to resolve?

PR #123, opened 2026-06-02
Card Implementation: Swap Top And Bottom

### LINKED ISSUE #113: Game Card-SwapTopAndBottom
Once played, the card goes to the discard pile. The user swaps the top and bottom card of the deck.

dongjoo224:


src/main/java/domain/Game.java
```diff
@@ -314,7 +314,19 @@ void applyClone() {
     }
 
     void applySwapTopAndBottom() {
```
jazlynho2028, line 341: Your current implementation removes the top and bottom cards and puts them both on top of the pile. Consider renaming addCard method in Deck to addCardToTop and add an addCardToBottom method in Deck. Make sure to rename all instances of addCard across any BVAs that mention it too

jazlynho2028 (changes requested):

src/main/java/domain/Game.java
dongjoo224: Made all the changes!

src/test/java/domain/DeckTests.java
```diff
@@ -924,4 +924,35 @@ private static Stream<Arguments> insertCardAtInvalidIndexCases() {
                         })
         );
     }
```
jazlynho2028, line 926: This is an integration test because you're creating new card objects to pass in. Make either one or both of the existing helpers in GameTests.java package private if you need card mocks. For these tests, I suggest mocking specific cards.

getCardMocksWithTypeExpectations
mockSpecificCard

src/test/java/domain/GameTests.java
```diff
@@ -1169,6 +1169,100 @@ public void advanceTurn_cannotEndTurn_failed() {
...
+
+		EasyMock.replay(players, drawPile, discardPile, turnManager);
+
+		Game game = new Game(players, drawPile, discardPile, turnManager);
+		game.applySwapTopAndBottom();
+
+		EasyMock.verify(drawPile);
+	}
+
+	@Test
```
jazlynho2028, line 1897: Same comment about this being an integration test

src/test/java/domain/GameTests.java
```diff
@@ -1169,6 +1169,100 @@ public void advanceTurn_cannotEndTurn_failed() {
...
+
+		Card card1 = new Card("SKIP_1", CardType.SKIP);
+		Card card4 = new Card("ATTACK_1", CardType.ATTACK);
+
+		final int drawPileSize = 4;
+		EasyMock.expect(drawPile.isEmpty()).andReturn(false);
+		EasyMock.expect(drawPile.size()).andReturn(drawPileSize);
+		EasyMock.expect(drawPile.removeTop()).andReturn(card1);
+		EasyMock.expect(drawPile.removeBottom()).andReturn(card4);
+
```
jazlynho2028, line 1911: You should verify that the correct card is passed into the addCardToTop vs. addCardToBottom. These tests pass even though the cards weren't actually swapped in the code

src/main/java/domain/Game.java
```diff
@@ -314,7 +314,19 @@ void applyClone() {
...
+            return;
+        }
+
+        if (drawPile.size() == 1) {
+            return;
+        }
+
+        Card top = drawPile.removeTop();
+        Card bottom = drawPile.removeBottom();
+
```
jazlynho2028, line 348: Is it just me or is this the same order as before? It should be drawPile.addCardToTop(bottom) and drawPile.addCardToBottom(top)

jazlynho2028 (changes requested):

src/main/java/domain/Game.java
dongjoo224: Yes, sorry about that! Fixed it.

src/test/java/domain/DeckTests.java
dongjoo224: Thanks for the catch! Made changes in my recent commits.

src/test/java/domain/GameTests.java
dongjoo224: Changed.

src/test/java/domain/GameTests.java
dongjoo224: Added this check to every test.

src/main/java/domain/Game.java
```diff
@@ -339,7 +339,19 @@ void applyClone() {
     }
 
     void applySwapTopAndBottom() {
-        // TODO
+        if (drawPile.isEmpty()) {
```
jazlynho2028: Two if statements can be condensed into one if (drawPile.size() <= 1)

src/test/java/domain/GameTests.java
```diff
@@ -1893,4 +1987,10 @@ private static boolean hasSameCardFields(Card card, CardType cardType, int idNum
 		return matchesType && matchesId;
 	}
 
+	private Card mockCardOfType(CardType cardType) {
```
jazlynho2028, line 1986: I like this method name. I made the same method in my branch, except I also replay the card so I can use it in a mockCardsOfTypes method. Would you mind adding replay here too?

jazlynho2028 (approved):
Approved with two minor changes

src/main/java/domain/Game.java
dongjoo224: Changed!

src/test/java/domain/GameTests.java
dongjoo224: Yep!

PR #124, opened 2026-06-02
Card Implementation: Super Skip

### LINKED ISSUE #110: Game Card-SuperSkip
When played this card goes to the discard pile. The user ends their turn immediately. If the user has to take multiple turns in a row, all of them are skipped.

Luke-Meyer-hue:
The Super Skip Card, when played allows for the turn to immediately go to the next player no matter how many turns the user has to take


src/main/java/domain/Game.java
```diff
@@ -302,7 +302,10 @@ void applyCatomicBomb() {
     }
 
     void applySuperSkip() {
```
jazlynho2028, line 390: This method can be two lines: 
turnManager.setDrawCount(0);
advanceTurn();

jazlynho2028 (changes requested):

src/main/java/domain/Game.java
Luke-Meyer-hue: changed it so that it simply sets the draw count

src/main/java/domain/Game.java
```diff
@@ -331,7 +331,8 @@ void applyCatomicBomb() {
...
-        // TODO
+        turnManager.setDrawCount(GameConstants.NUM_SUPER_SKIP_DRAW_COUNT);
```
jazlynho2028: You can just do turnManager.setDrawCount(0) here. You don't need to follow clean code if following it makes the code harder to read.

src/main/java/ui/PlayerDeckController.java
```diff
@@ -207,6 +207,9 @@ void onPlayCardsButton() {
                 case SKIP:
                     handleChangeCurrentPlayer(model.getCurrentPlayerIndex());
                     break;
+                case SUPER_SKIP:
```
jazlynho2028: I believe SKIP should also do renderNextTurn because the draw pile needs to be updated. So you can remove these two lines below to let the case fall through to SUPER_SKIP

handleChangeCurrentPlayer(model.getCurrentPlayerIndex());
break;

renderNextTurn still fetches the currentPlayerIndex from the model, so it won't render the next player if drawCount is not 0.

jazlynho2028 (changes requested):

src/main/java/domain/Game.java
Luke-Meyer-hue: Sounds good, I'll take it out in that case

src/main/java/ui/PlayerDeckController.java
Luke-Meyer-hue: Ok, the logic in renderNextTurn uses handleChangeCurrentPlayer anyways so SKIP goes through the same process anyways. I'll change this.

jazlynho2028 (approved):

PR #125, opened 2026-06-02
Card Implementation: Recycle

### LINKED ISSUE #118: Game OriginalCard-Recycle
When played, this card goes to the discard pile. Before it enters the discard pile, the user shuffles the discard pile and draws from the bottom of it. This card ends their turn.

dongjoo224:
The Recycle card when drawn shuffles the discard pile then draws the bottom card to immediately end one turn.


src/main/java/domain/Game.java
```diff
@@ -335,7 +335,17 @@ void applyRagebait() {
     }
 
     void applyRecycle() {
```
jazlynho2028: Keep in mind that this is the backend. We have a GUI, so we don't draw cards for the player. applyRecycle should update conditions that will allow the player to draw from the discard pile. Right now, Game has a canDrawFromDiscard method that always returns false. Consider storing a boolean in Game for that and set it to true to conditionally enable the discard pile to be drawn. Then you'd disable the draw pile in onPlayCardsButton and re enable it in onDiscardPile controller method

jazlynho2028 (changes requested):

src/main/java/domain/Game.java
dongjoo224: I understand that applyRecycle() should just set the variable and not draw the card itself. Then for onDiscardPile() in the controller, when the player clicks the discard pile button the controller needs to actually draw from discard pile so is it okay for a drawFromDiscard() method in Game to draw at that point since it's triggered by a button click?

jazlynho2028: Yes that's exactly what should happen.

jazlynho2028: Recycle should re render the discard pile after shuffle (call rebindHandCards()), and force the player to be in draw mode by disabling the hand cards (canDraw = isGameOngoing && turnManager.drawCount > 0 && canDrawFromDiscard()). So you might have to make canDrawFromDiscard a stored boolean var and rename the current method to getCanDrawFromDiscard. If you think of a cleaner way to implement this, let me know.

dongjoo224: Got it. I implemented this in my newest commits with some changes.

jazlynho2028 (changes requested):
If you're automatically drawing the card, I think you don't need the canDrawFromDiscard variable anymore since the UI should always render the discard pile as disabled. You can delete that parameter in the controller and views too.

jazlynho2028:
I could be wrong though. Let me know if it's actually needed.

jazlynho2028:
https://github.com/nu-cs-sqe/course-project-20252603-team-14-20252603/pull/139#pullrequestreview-4445869769

Irene's Draw from the bottom also automatically draws a card. I wrote down how to implement these types of draw cards in the above review. It should also apply to your Recycle card. Hope that helps!

Also for Recycle, keep in mind that if the discard pile is initially empty, then once Recycle is played, the discard pile will contain that Recycle card, so playing Recycle would just add Recycle back to the hand (although this should already happen given the current code in main).

src/main/java/ui/PlayerDeckController.java
```diff
@@ -241,14 +240,25 @@ void updateByCardType(CardType cardType) {
...
                 view.renderTurnControlSection(false, false);
+                break;
+            case RECYCLE:
+                Card drawnCard = model.drawFromRecycle();
+                if (drawnCard.getType() == CardType.EXPLODING_KITTEN) {
+                    handleDrawExplodingKitten(drawnCard.getId());
+                }
+                else {
+                    rebindHandCards();
+                    updateDrawPile();
```
jazlynho2028: Since the onDrawPile controller method is = exactly the same as this RECYCLE case code except for model.drawFromRecycle instead of model.drawFromPile, you can make a private helper for the if else block.

Recycle technically would never draw an exploding kitten from the discard pile because exploding kittens are always returned to the deck or removed from the game. But it's good to handle that case too just in case.

jazlynho2028: you don't technically need updateDrawPile since you're drawing from discard, but calling it anyways doesn't hurt because the controller always fetches the latest model data in the update methods.

src/main/java/domain/Game.java
```diff
@@ -464,7 +462,11 @@ void applyRagebait() {
...
```
jazlynho2028: It would be cleaner to combine applyRecycle and drawFromRecycle into a drawRecycle method and just remove the RECYCLE case in applyByCardType

```
public Card drawRecycle() {
    discardPile.shuffle();
    return drawCard(discardPile::peekBottom, discardPile::removeBottom);
}
```

src/main/java/ui/PlayerDeckController.java
```diff
@@ -241,14 +240,25 @@ void updateByCardType(CardType cardType) {
                 view.renderPlayerNameTags(model.getCurrentPlayerIndex(), false,
                         model.getDeadIndices());
                 view.renderTurnControlSection(false, false);
+                break;
+            case RECYCLE:
+                Card drawnCard = model.drawFromRecycle();
```
jazlynho2028: You'll need to add an updateDiscardPile after this line if you combine the apply and draw recycle methods

src/main/java/ui/PlayerDeckController.java
```diff
@@ -241,14 +240,25 @@ void updateByCardType(CardType cardType) {
...
+                }
                 break;
 			default:
                 break;
         }
     }
 
     private void updateDiscardPile() {
-        view.renderDiscardPile(model.canDrawFromDiscard(), model.getTopDiscardId());
+        view.renderDiscardPile(false, model.getTopDiscardId());
```
jazlynho2028: In view, you can remove the canDraw parameter from view.renderDiscardPile and just always disable the discardPileButton by removing discardPileButton.setDisable(!canDraw) from renderDiscardPile, and adding a discardPileButton.setDisable(true) in buildDiscardPileButton()

^ Since the discard pile will never be clickable.

jazlynho2028: update bvas and tests as necessary

jazlynho2028 (changes requested):
The idea is good, just some refactoring for cleaner code.

src/main/java/ui/PlayerDeckController.java
dongjoo224: Got it!

src/main/java/domain/Game.java
dongjoo224: Yes, good idea. I'll make this change.

src/main/java/ui/PlayerDeckController.java
dongjoo224: Done!

src/main/java/ui/ExplodingKittensApp.java
```diff
@@ -81,6 +81,8 @@ private void initializeGame(PlayerCreateController createController, Stage stage
         Game model = new Game(players, drawPile, discardPile, turnManager);
         model.setUp();
 
+        players.get(0).addCardToHand(new Card("RECYCLE_1", CardType.RECYCLE));
```
jazlynho2028: I'm assuming this was for testing purposes, but can you remove it before merging into main?

src/main/java/ui/PlayerDeckController.java
```diff
@@ -241,14 +232,30 @@ void updateByCardType(CardType cardType) {
                 view.renderPlayerNameTags(model.getCurrentPlayerIndex(), false,
                         model.getDeadIndices());
                 view.renderTurnControlSection(false, false);
+                break;
+            case RECYCLE:
+                updateDiscardPile();
```
jazlynho2028: I made some significant refactoring in controller in my winner winner branch. It makes updating the UI less tedious. After it gets merged into main, you would only have to call the updateDiscardPile once after drawRecycle.

src/main/java/ui/PlayerDeckController.java
```diff
@@ -241,14 +232,30 @@ void updateByCardType(CardType cardType) {
...
+                handleDrawnCard(model.drawRecycle());
```
jazlynho2028, line 249: Good

jazlynho2028 (changes requested):
Looks good! Just minor clean-up points. Would you mind letting me merge my winner winner branch into main first? It has some changes to the controller that would make the recycle ui update slightly cleaner too.

src/main/java/ui/ExplodingKittensApp.java
dongjoo224: Yes! Forgot about that, thanks for the reminder.

src/main/java/ui/PlayerDeckController.java
dongjoo224: Got it. Yeah, it felt weird calling update twice. I'll make these changes once the Winner Winner card is merged.

src/main/java/ui/PlayerDeckController.java
```diff
@@ -230,6 +230,15 @@ void onPlayCardsButton() {
         });
     }
 
+    private void handleDrawnCard(Card drawnCard) {
```
jazlynho2028, line 226: you can make onDrawPile use this method too

jazlynho2028 (approved):
Just one small change. Looks good otherwise!

src/main/java/ui/PlayerDeckController.java
dongjoo224: True. Will make this change.

PR #126, opened 2026-06-03
Card Implementation: Double Up

### LINKED ISSUE #119: Game OriginalCard-DoubleUp
Once played, this card goes to the discard pile. The user draws two cards instead of one at the end of their turn.

ejeankuo:


src/main/java/domain/Game.java
```diff
@@ -339,7 +339,13 @@ void applyRecycle() {
     }
```
jazlynho2028, line 519: you don't need to draw the card for the player because we have a GUI. all this method needs to do is increment the draw count. So just create the following method in TurnManager.java and call it in applyDoubleUp 

public void incrementDrawCount() {
        drawCount++
    }

jazlynho2028 (changes requested):

src/main/java/domain/Game.java
ejeankuo: Done!

jazlynho2028 (approved):
Is there a case where the draw pile is disabled and playing double up causes it to enable?

I can't think of any cases where that would happen, but if you do, then you'd need to add a case in the onPlayCardsButton switch statement for case DOUBLE_UP and call updateDrawPile() in it. Otherwise this looks good.

jazlynho2028 (changes requested):
Do you think double up should automatically draw the card for the player like draw from the bottom and recycle? I'm worried it could interfere with Attack if they are played together. For example, a player could play Double Up then Attack, and the next player would have to draw three cards instead of two.

jazlynho2028:
> Do you think double up should automatically draw the card for the player like draw from the bottom and recycle? I'm worried it could interfere with Attack if they are played together. For example, a player could play Double Up then Attack, and the next player would have to draw three cards instead of two.

Or maybe this could be considered a strategy

ejeankuo:
> > Do you think double up should automatically draw the card for the player like draw from the bottom and recycle? I'm worried it could interfere with Attack if they are played together. For example, a player could play Double Up then Attack, and the next player would have to draw three cards instead of two.
> 
> Or maybe this could be considered a strategy

Resolved: We decided to keep this as is so that Double Up (usually a penalty for the person who drew it) can be turned into a stacked attack towards the next player (now a strategy)

Luke-Meyer-hue (approved):
The code looks good, there is only one test as the code is incredibly simple, just calling increment count, and as increment count has already been tested itself, there isn't much to test here.

jazlynho2028 (approved):
Looks good

PR #127, opened 2026-06-03
Card Implementation: Godcat

### LINKED ISSUE #111: Game Card-Godcat
Once played, this card is removed from the game (not in discard pile or draw pile). Play any card of your choice, similar to a multicolor in uno.

dongjoo224:


src/main/java/ui/PlayerDeckView.java
```diff
@@ -766,7 +793,75 @@ private StackPane buildOverlayLayer() {
...
+
+        godcatConfirmButton.setText("Confirm");
+        godcatConfirmButton.setDisable(true);
+        godcatConfirmButton.getStyleClass().addAll("turn-control-button", "h5");
+
+        overlayContent.getChildren().addAll(title, cardScrollPane, godcatConfirmButton);
+        godcatOverlay.getChildren().add(overlayContent);
+    }
+
+    public void showGodcatOverlay() {
```
jazlynho2028: The way I did overlay for exploding kitten is I define one overlayLayer. Your implementation works too but having one overlayLayer would reduce the hassle of define multiple overlay layers for each card that needs a UI

public void buildExplodeOverlay(boolean hasDefuse, String explodingCardId) {
        VBox content = new VBox();
        ....populate content with buttons

        overlayLayer.getChildren().setAll(content);
        showOverlay();
    }

Then hideOverlay does

public void hideOverlay() {
        overlayLayer.setVisible(false);
        overlayLayer.setMouseTransparent(true);
        overlayLayer.getChildren().clear();
    }

Then in controller:

void onExplodeButton() {
        attempt(onError, () -> {
            model.applyExplode();

            view.hideOverlay();

            // UI update to advance turn
            handleChangeCurrentPlayer(model.getCurrentPlayerIndex());
            updateDrawPile();
            updateTurnControls();
        });
    }

jazlynho2028 (changes requested):

src/main/java/ui/PlayerDeckView.java
dongjoo224: That's a good point. Ok, I refactored my implementation to use a shared overlayLayer field in the constructor instead of a separate godcatOverlay, and that's now configured in buildOverlayLayer() which changed to a void method. And I made the changes in buildGodcatOverlay() to call overlayLayer.getChildren()..., showOverlay(), and hideOverlay().

src/main/java/ui/PlayerDeckView.java
```diff
@@ -762,11 +788,87 @@ private void renderTurnControlButton(Button turnControlButton, String label) {
         );
     }
 
-    private StackPane buildOverlayLayer() {
-        StackPane overlayLayer = new StackPane();
+    private void buildOverlayLayer() {
```
jazlynho2028, line 775: private void buildOverlayLayer() {
        overlayLayer.setAlignment(Pos.CENTER);
        overlayLayer.getStyleClass().add("overlay");
        overlayLayer.setVisible(false);
        overlayLayer.setMouseTransparent(true);
}

I made the following changes in styles.css that you can use for your overlay too
in root
gray-6-tr: #01010190;

.overlay {
    -fx-background-color: gray-6-tr;
}

.overlay-content {
    -fx-spacing: 16px;
}

.overlay-button {
    -fx-background-radius: 8;
    -fx-border-width: 2;
    -fx-border-radius: 16;
    -fx-pref-width: 176px;
    -fx-padding: -2 0 -2 0;
}

.overlay-button.defuse {
    -fx-background-color: linear-gradient(to bottom, yellow-3, defuse, recycle);
    -fx-text-fill: yellow-0;
}

.overlay-button.explode {
    -fx-background-color: linear-gradient(to bottom, yellow-3, red-3, orange-3);
    -fx-text-fill: magenta-6;
}

Either define a .overlay-button.confirm class in css for your button and copy over the styling from another button, or rename "turn-control-button" css class to "button-secondary "or something and change all instances of "turn-control-button" to match

src/main/java/ui/PlayerDeckView.java
```diff
@@ -762,11 +788,87 @@ private void renderTurnControlButton(Button turnControlButton, String label) {
...
+        overlayLayer.setVisible(false);
+        overlayLayer.setMouseTransparent(true);
         overlayLayer.setPickOnBounds(false);
+    }
+
+    public void buildGodcatOverlay(List<CardType> cardTypes) {
+        VBox overlayContent = new VBox();
+        overlayContent.setAlignment(Pos.CENTER);
+        overlayContent.setSpacing(UIConstants.GODCAT_SCROLLPANE_PADDING);
+        overlayContent.setStyle("-fx-background-color: rgba(0, 0, 0, 0.7); -fx-padding: 20;");
```
jazlynho2028: use .overlay-content css style class above. Keep styling in css so the view doesn't get cluttered

src/main/java/ui/PlayerDeckView.java
```diff
@@ -762,11 +788,87 @@ private void renderTurnControlButton(Button turnControlButton, String label) {
...
+    private void buildOverlayLayer() {
+        overlayLayer.setVisible(false);
+        overlayLayer.setMouseTransparent(true);
         overlayLayer.setPickOnBounds(false);
+    }
+
+    public void buildGodcatOverlay(List<CardType> cardTypes) {
+        VBox overlayContent = new VBox();
+        overlayContent.setAlignment(Pos.CENTER);
+        overlayContent.setSpacing(UIConstants.GODCAT_SCROLLPANE_PADDING);
```
jazlynho2028: Spacing can also be set in css. Feel free to add to the existing overlay-content css class. You can add your spacing there. Keep setAlignment and setPrefSize in java code.

src/main/java/ui/PlayerDeckView.java
```diff
@@ -762,11 +788,87 @@ private void renderTurnControlButton(Button turnControlButton, String label) {
...
+        overlayContent.setStyle("-fx-background-color: rgba(0, 0, 0, 0.7); -fx-padding: 20;");
+        overlayContent.setPrefSize(UIConstants.SCENE_WIDTH, UIConstants.SCENE_HEIGHT);
+
+        Text title = new Text("Choose a card for Godcat");
+        title.getStyleClass().add("h3");
+        title.setFill(javafx.scene.paint.Color.WHITE);
```
jazlynho2028: You can do title.getStyleClass().add("caption"). That css class already has font color and size defined

src/main/java/ui/PlayerDeckView.java
```diff
@@ -762,11 +788,87 @@ private void renderTurnControlButton(Button turnControlButton, String label) {
...
+    }
+
+    public void buildGodcatOverlay(List<CardType> cardTypes) {
+        VBox overlayContent = new VBox();
+        overlayContent.setAlignment(Pos.CENTER);
+        overlayContent.setSpacing(UIConstants.GODCAT_SCROLLPANE_PADDING);
+        overlayContent.setStyle("-fx-background-color: rgba(0, 0, 0, 0.7); -fx-padding: 20;");
+        overlayContent.setPrefSize(UIConstants.SCENE_WIDTH, UIConstants.SCENE_HEIGHT);
+
+        Text title = new Text("Choose a card for Godcat");
```
jazlynho2028: remember to add any strings to .properties and use assetProvider.getString("playerDeckScreen.godcatCaption")

src/main/java/ui/PlayerDeckView.java
jazlynho2028: you can remove

title.getStyleClass().add("h3");
title.setFill(javafx.scene.paint.Color.WHITE);

after that

src/main/java/ui/PlayerDeckView.java
jazlynho2028: Last part refers to your godcatConfirmButton.getStyleClass().addAll("turn-control-button", "h5");

src/main/java/ui/PlayerDeckView.java
```diff
@@ -762,11 +788,87 @@ private void renderTurnControlButton(Button turnControlButton, String label) {
...
 
-    private StackPane buildOverlayLayer() {
-        StackPane overlayLayer = new StackPane();
+    private void buildOverlayLayer() {
+        overlayLayer.setVisible(false);
+        overlayLayer.setMouseTransparent(true);
         overlayLayer.setPickOnBounds(false);
+    }
+
+    public void buildGodcatOverlay(List<CardType> cardTypes) {
```
jazlynho2028: Ideally you should make private helpers to make this method smaller. Also would be appropriate to name this buildCardSelectOverlay since other card effects will use it

jazlynho2028 (changes requested):

src/main/java/ui/PlayerDeckView.java
jazlynho2028: Or if you want to keep the h3 font and size, make a new .overlay-title css class to change the color and add "h3" and "overlay-title" to the title style class. When using colors in css file, only use the pre-defined color variables in .root at the top of the css file

src/main/java/ui/PlayerDeckView.java
```diff
@@ -762,11 +788,87 @@ private void renderTurnControlButton(Button turnControlButton, String label) {
...
+            cardButton.setOnMouseClicked(e -> {
+                selectedGodcatCardType = cardType;
+                cardOptions.getChildren().forEach(n ->
+                        ((ToggleButton) n).setSelected(false));
+                cardButton.setSelected(true);
+                godcatConfirmButton.setDisable(false);
+            });
+            cardOptions.getChildren().add(cardButton);
+        }
+
```
jazlynho2028: You can rename the existing buildHandScrollPane method to buildCardScrollPane and refactor it to take in an HBox card

private ScrollPane buildCardScrollPane(HBox content) {

        ScrollPane handScrollPane = new ScrollPane(content);
        handScrollPane.setHbarPolicy(ScrollPane.ScrollBarPolicy.ALWAYS);
        handScrollPane.setVbarPolicy(ScrollPane.ScrollBarPolicy.NEVER);
        handScrollPane.getStyleClass().add("scroll-pane");

        content.setAlignment(Pos.CENTER);
        content.setMinWidth(UIConstants.SCENE_WIDTH);
        content.getStyleClass().add("hand-cards-container");

        return handScrollPane;

    }

I removed the private helper renderCardScrollPaneContent 

Then the handScrollPane build line in buildPlayerHandSection becomes 
ScrollPane handScrollPane = buildCardScrollPane(handCardsContainer);

Then you can call buldCardScrollPane(cardOptions) in your godcat overlay

src/main/java/ui/PlayerDeckView.java
```diff
@@ -762,11 +788,87 @@ private void renderTurnControlButton(Button turnControlButton, String label) {
...
+        Text title = new Text("Choose a card for Godcat");
+        title.getStyleClass().add("h3");
+        title.setFill(javafx.scene.paint.Color.WHITE);
+
+        HBox cardOptions = new HBox(UIConstants.GODCAT_CARD_SPACING);
+        cardOptions.setAlignment(Pos.CENTER);
+
+        for (CardType cardType : cardTypes) {
+            String cardId = cardType.name().replace("_", "") + "_1";
```
jazlynho2028: There is a static DeckBuilder method that takes in a cardType and number to create a cardId already

public static String createCardId(CardType type, int num) {

        if (num <= 0) {
            throw new IllegalArgumentException("error.invalidCardIDNumber");
        }

        String cardTypeNameWithoutUnderscore = type.name().replace("_", "");

        return String.format(
                "%s_%d",
                cardTypeNameWithoutUnderscore,
                num
        );
    }

jazlynho2028 (changes requested):

src/main/java/ui/PlayerDeckView.java
```diff
@@ -762,11 +788,87 @@ private void renderTurnControlButton(Button turnControlButton, String label) {
...
+        overlayLayer.setMouseTransparent(true);
         overlayLayer.setPickOnBounds(false);
+    }
+
+    public void buildGodcatOverlay(List<CardType> cardTypes) {
+        VBox overlayContent = new VBox();
+        overlayContent.setAlignment(Pos.CENTER);
+        overlayContent.setSpacing(UIConstants.GODCAT_SCROLLPANE_PADDING);
+        overlayContent.setStyle("-fx-background-color: rgba(0, 0, 0, 0.7); -fx-padding: 20;");
+        overlayContent.setPrefSize(UIConstants.SCENE_WIDTH, UIConstants.SCENE_HEIGHT);
```
jazlynho2028: I'm not sure if this line does anything because it didn't work for me to center my content when I did this. It might be different for you though cuz your scroll pane extends the whole width of the screen. Does the formatting change if you remove it? 

You can try the following if you find that the line is not actually doing anything:
overlayContent.setMaxSize(Region.USE_PREF_SIZE, Region.USE_PREF_SIZE);

src/main/java/ui/PlayerDeckView.java
dongjoo224: Renamed to "button-secondary" and changed the styles.css as suggested!

src/main/java/ui/PlayerDeckView.java
dongjoo224: Got it. I separated the method into helpers that can be reused for other cards.

src/main/java/ui/PlayerDeckView.java
dongjoo224: Made changes!

src/main/java/ui/PlayerDeckView.java
dongjoo224: Removed.

src/main/java/ui/PlayerDeckView.java
dongjoo224: Removed it and it seems like it's not affecting the overlay so I'll replace it with the line you suggested.

src/main/java/ui/PlayerDeckView.java
dongjoo224: Thanks for the reminder!

src/main/java/ui/PlayerDeckView.java
dongjoo224: Decided to make a new overlay-title to styles.css.

src/main/java/ui/PlayerDeckView.java
dongjoo224: Oh nice, I'll use that.

src/main/java/ui/PlayerDeckView.java
dongjoo224: Done!

src/main/java/ui/PlayerDeckView.java
```diff
@@ -36,6 +38,28 @@ public class PlayerDeckView {
     private final StackPane root;
     private final HBox turnControlSection;
 
+    private final StackPane overlayLayer;
+    public final Button godcatConfirmButton;
+    private CardType selectedGodcatCardType;
+
+    private static final List<CardType> GODCAT_CARD_OPTIONS = List.of(
```
jazlynho2028: View should be dumb. You should put this in controller instead and pass the list into view.showCardSelectOverlay. Then you don't need view.showGodcatOverlay().

src/main/java/ui/PlayerDeckView.java
```diff
@@ -757,16 +785,103 @@ private void buildAndAddTurnControlButtonsBeforeGameStart() {
...
+        ToggleButton cardButton = new ToggleButton();
+        cardButton.getStyleClass().addAll("card", "front");
+        cardButton.setContentDisplay(ContentDisplay.GRAPHIC_ONLY);
+
+        VBox cardFront = buildCardFront(cardId);
+        cardButton.setGraphic(cardFront);
+
+        cardButton.setOnMouseClicked(e -> {
+            selectedGodcatCardType = cardType;
+            cardOptions.getChildren().forEach(n ->
```
jazlynho2028: rename n

src/main/java/ui/PlayerDeckView.java
```diff
@@ -50,6 +74,9 @@ public PlayerDeckView(AssetProvider assetProvider) {
         playCardsButton = new Button();
         endTurnButton = new Button();
         turnControlSection = new HBox();
+        overlayLayer = new StackPane();
+        godcatConfirmButton = new Button();
+        selectedGodcatCardType = null;
```
jazlynho2028: Don't initialize anything as null. Put a default CardType.ATTACK or something.

jazlynho2028 (changes requested):
Thanks for the big refactors! Just some minor changes.

src/main/java/ui/PlayerDeckView.java
dongjoo224: Changed!

src/main/java/ui/PlayerDeckView.java
dongjoo224: True. Changed !

src/main/java/ui/PlayerDeckView.java
dongjoo224: Renamed

jazlynho2028 (changes requested):
Seems to have an error for the following test case 
public void onPlayCardsButton_godcatPlayed_overlayShown

jazlynho2028:
Either make GODCAT_CARD_OPTIONS package private to access in tests or perhaps putting it in GameConstants is a better move.

jazlynho2028:
I can make the changes.

jazlynho2028 (approved):

PR #129, opened 2026-06-03
Card Implementation: Attack

### LINKED ISSUE #103: Game Card-Attack
The player that plays this card ends their turn immediately, then the next player is forced to play two normal turns in which they can play any cards and then draw.

ejeankuo:
This is an updated PR from PR #128 with cleaned up commit history, isolated to applyAttack(). PR #128 mistakenly was branched from PR #126 and contained commit history from PR #126, which I cleaned up in this current PR for readability purposes.


jazlynho2028 (changes requested):
Offered a cleaner way to implement attack

src/main/java/domain/Game.java
```diff
@@ -282,7 +282,16 @@ void setIsFaceUp(boolean isFaceUp) {
     }
 
     void applyAttack() {
```
jazlynho2028, line 384: This can be reduced to two lines:

turnManager.incrementTurn();

jazlynho2028, line 384: It might be better to refactor canEndTurn and store it as a boolean variable so you can also set canEndTurn to true for the UI to update properly. You'd have to make sure the canEndTurn gets updated correctly for any existing code though

jazlynho2028 (changes requested):

src/main/java/domain/Game.java
ejeankuo: I think reducing it to these lines breaks the logic for handling stacked Attack cards, right? Players can stack Attacks and increase the drawCount up to 6, which is why I implemented logic for setting drawCount to be the currentDrawCount + 2 before advancing the turn.

ejeankuo: I like this idea for making sure UI aligns with the backend logic, I'll look into this change.

jazlynho2028: Resolved. Decided to try using an incrementDrawCount method in TurnManager

jazlynho2028 (changes requested):
https://github.com/nu-cs-sqe/course-project-20252603-team-14-20252603/pull/133#discussion_r3366872532

See conversation with Luke about implementing Attack/Targeted Attack. You should discuss with Luke since he is doing targeted attack.

jazlynho2028 (changes requested):
Attack looks good! If canPlay has the same logic as canDraw, can you just keep canDraw and replace everything that uses canPlay or getCanPlay with canDraw?

jazlynho2028 (approved):

PR #132, opened 2026-06-05
Card Implementation: See The Future

### LINKED ISSUE #106: Game Card-SeeTheFuture
When played, the card goes to the discard pile and the player that played the see the future is allowed to look at the top three cards of the deck and then they will be put back in the same order.

jazlynho2028:
Includes: 

- See The Future UI and domain implementation
- PlayerDeckController test refactoring to use helper expectation methods
- Defuse implementation to account for Godcat and Clone as defusable cards


Luke-Meyer-hue (approved):
Everything looks good to me, creating the see the future overalys required as well as the functions and the tests. Additionally, added a few other changes like isDefusable()

PR #133, opened 2026-06-06
Card Implemented: Targeted Attack

### LINKED ISSUE #115: Game Card-TargetedAttack
Once played, this card goes to the discard pile. The user's turn ends immediately and forces a targeted player to take two turns in a row. Play resumes from that player.

Luke-Meyer-hue:
Creates the card for targeted attack. For those reviewing, if you would take a look at src/main/java/ui/PlayerDeckController.java (the method onNameTag) the way that I allow the controller to start the targeted requires using null, and I wasn't sure if there was a better way.

### DIFF
--- a/docs/bva/TurnManager.md
+++ b/docs/bva/TurnManager.md
@@ -39,7 +39,7 @@
   - **Expected output**: drawCount = 1
 
 ### Method under test: `incrementTurn()`
-- **TC7: Advance turn away from starting baseline across varying player sizes** ( :x: )
+- **TC7: Advance turn away from starting baseline across varying player sizes** ( :white_check_mark: )
   - **Name of the test**: incrementTurn_boundaryScenarios_updatesPlayerIndexCorrectly
   - **State of the system**: 
     - numPlayers = N (N = 2, 3, 4)
--- a/docs/bva/TurnManager.md
+++ b/docs/bva/TurnManager.md
@@ -50,7 +50,7 @@
     - roundCount = 1
 
 
-- **TC8: Advance turn on boundary just before list wraps around** ( :x: )
+- **TC8: Advance turn on boundary just before list wraps around** ( :white_check_mark: )
   - **Name of the test**: incrementTurn_boundaryScenarios_updatesPlayerIndexCorrectly
   - **State of the system**: 
     - numPlayers = N (N = 2, 3, 4)
--- a/docs/bva/TurnManager.md
+++ b/docs/bva/TurnManager.md
@@ -39,41 +39,36 @@
   - **Expected output**: drawCount = 1
 
 ### Method under test: `incrementTurn()`
-- **TC7: Advance turn away from starting baseline across varying player sizes** ( :white_check_mark: )
+- **TC7: Advance turn away from starting baseline across varying player sizes** ( :x: )
   - **Name of the test**: incrementTurn_boundaryScenarios_updatesPlayerIndexCorrectly
   - **State of the system**: 
     - numPlayers = N (N = 2, 3, 4)
     - currentPlayerIndex = 0
     - roundCount = 1
-    - drawCount = 0
   - **Expected output**: 
     - currentPlayerIndex = 1
     - roundCount = 1
-    - drawCount = 1
 
-- **TC8: Advance turn on boundary just before list wraps around** ( :white_check_mark: )
+
+- **TC8: Advance turn on boundary just before list wraps around** ( :x: )
   - **Name of the test**: incrementTurn_boundaryScenarios_updatesPlayerIndexCorrectly
   - **State of the system**: 
     - numPlayers = N (N = 2, 3, 4)
     - currentPlayerIndex = N - 2
     - roundCount = 2
-    - drawCount = 1
   - **Expected output**: 
     - currentPlayerIndex = N - 1
     - roundCount = 2
-    - drawCount = 2
 
-- **TC9: Advance that hits upper boundary, wrapping index back to zero** ( :white_check_mark: )
+- **TC9: Advance that hits upper boundary, wrapping index back to zero** ( :x: )
   - **Name of the test**: incrementTurn_boundaryScenarios_updatesPlayerIndexCorrectly
   - **State of the system**: 
     - numPlayers = N (N = 2, 3, 4)
     - currentPlayerIndex = N - 1
     - roundCount = 1
-    - drawCount = 0
   - **Expected output**: 
     - currentPlayerIndex = 0
     - roundCount = 2
-    - drawCount = 1
 
 ### Method under test: `setCurrentPlayerIndex(int newPlayerIndex)`
 - **TC10: Negative index** ( :white_check_mark: )
--- a/docs/bva/TurnManager.md
+++ b/docs/bva/TurnManager.md
@@ -92,12 +92,12 @@
   - **Expected output**: throw IllegalArgumentException "error.invalidPlayerIndex"
 
 ### Method under test: `incrementDrawCount()`
-- **TC14: Decrementing from zero throws state exception** ( :x: )
-  - **Name of the test**: incrementDrawCount_fromZero_returnsOne
+- **TC14: Decrementing from zero throws state exception** ( :white_check_mark: )
+  - **Name of the test**: incrementDrawCount_validDrawCount_incrementedByOne
   - **State of the system**: drawCount = 0
   - **Expected output**: drawCount = 1
 
 - **TC15: incrementing from a positive value successfully increases count** ( :x: )
-  - **Name of the test**: incrementDrawCount_positiveDrawCount_incrementedByOne
+  - **Name of the test**: incrementDrawCount_validDrawCount_incrementedByOne
   - **State of the system**: drawCount = 1
   - **Expected output**: drawCount = 2
--- a/docs/bva/TurnManager.md
+++ b/docs/bva/TurnManager.md
@@ -23,7 +23,7 @@
     - roundCount = 1
 
 ### Method under test: `decrementDrawCount()`
-- **TC4: Decrementing from zero throws state exception*** ( :white_check_mark: )
+- **TC4: Decrementing from zero throws state exception** ( :white_check_mark: )
   - **Name of the test**: decrementDrawCount_fromZero_throwsIllegalStateException
   - **State of the system**: drawCount = 0
   - **Expected output**: IllegalStateException("error.negativeDrawCount")
@@ -90,3 +90,14 @@
   - **Name of the test**: setCurrentPlayerIndex_invalidIndex_failed
   - **State of the system**: TurnManager constructed with 1 player, newPlayerIndex = 1
   - **Expected output**: throw IllegalArgumentException "error.invalidPlayerIndex"
+
+### Method under test: `incrementDrawCount()`
+- **TC14: Decrementing from zero throws state exception** ( :x: )
+  - **Name of the test**: incrementDrawCount_fromZero_returnsOne
+  - **State of the system**: drawCount = 0
+  - **Expected output**: drawCount = 1
+
+- **TC15: incrementing from a positive value successfully increases count** ( :x: )
+  - **Name of the test**: incrementDrawCount_positiveDrawCount_incrementedByOne
+  - **State of the system**: drawCount = 1
+  - **Expected output**: drawCount = 2
--- a/docs/bva/TurnManager.md
+++ b/docs/bva/TurnManager.md
@@ -60,7 +60,7 @@
     - currentPlayerIndex = N - 1
     - roundCount = 2
 
-- **TC9: Advance that hits upper boundary, wrapping index back to zero** ( :x: )
+- **TC9: Advance that hits upper boundary, wrapping index back to zero** ( :white_check_mark: )
   - **Name of the test**: incrementTurn_boundaryScenarios_updatesPlayerIndexCorrectly
   - **State of the system**: 
     - numPlayers = N (N = 2, 3, 4)
--- a/docs/bva/TurnManager.md
+++ b/docs/bva/TurnManager.md
@@ -97,7 +97,7 @@
   - **State of the system**: drawCount = 0
   - **Expected output**: drawCount = 1
 
-- **TC15: incrementing from a positive value successfully increases count** ( :x: )
+- **TC15: incrementing from a positive value successfully increases count** ( :white_check_mark: )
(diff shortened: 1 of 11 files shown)


src/main/java/domain/Game.java
```diff
@@ -375,23 +339,17 @@ void applyClone() {
...
 
     void applyDrawFromTheBottom() {
         // TODO
     }
 
-    void applyTargetedAttack() {
-        // TODO
+    public void applyTargetedAttack(int targetPlayerIndex) {
+        getCurrentPlayer().deselectHandCards();
+        turnManager.setCurrentPlayerIndex(targetPlayerIndex);
```
jazlynho2028: This doesn't account for incrementing rounds properly or handling stacking attacks. Since Ejean is doing Attack, you should discuss with her how to do this and decide if turnManager should be refactored and/or needs new methods.

jazlynho2028 (changes requested):

src/main/java/domain/Game.java
jazlynho2028:
```
void applyAttack() {
        turnManager.incrementTurn();

        if (turnManager.getDrawCount() > 1) {
            turnManager.incrementDrawCount();
        }
    }

```
This is what Ejean has in her latest commit to remote. I don't think this is correct but either way it doesn't translate directly with targeted attack in terms of handling rounds and incrementing draw count properly.

I think a good fix may be to refactor turnManager (+ revise tests and bvas) so that turnManager.incrementTurn() doesn't increment the drawCount. Instead, Game calls incrementDrawCount() separately. 

Then Game would have
```
public void advanceTurn() {
        if (!canEndTurn()) {
            throw new IllegalStateException("error.cannotEndTurn");
        }
        turnManager.incrementTurn();
        drawCount = 1;
    }
```

and applyAttack would be
```
void applyAttack() {
        turnManager.incrementTurn();
        
        updateDrawCountForAttack();
}
```
and applyTargetedAttack would be 
```
void applyTargetedAttack(int targetPlayerIndex) {

        while (turnManager.getCurrentPlayerIndex != targetPlayerIndex) {
                turnManager.incrementTurn();
        }
        
        updateDrawCountForAttack();
}
```
and

```
private void updateDrawCountForAttack() {
        turnManager.incrementDrawCount();
        
        if (turnManager.getDrawCount() > 2) {
                while (int i = 0; i < ATTACK_NUM_DRAW; i++) {
                       turnManager.incrementDrawCount(); 
                }
        }
}
```

Then you can make a private helper for the if statement and use it for both. You should also discuss with Ejean to see if there's a better way.

Luke-Meyer-hue: I see, I hadn't realized that targeted attack stacks like that, thank you for notifying me. I'll split appart incrementDrawCount from incrementTurn, I don't believe we need a for loop because just getting the draw count and adding two should suffice, but otherwise, I completely agree with your adjustments. I have added them now.

src/main/java/domain/GameConstants.java
```diff
@@ -12,12 +12,16 @@ private GameConstants() { }
...
+	public static final int NUM_DEFUSES_IN_GAME = 5;
+
 	public static final int ONE_CARD = 1;
 	public static final int TWO_CARDS = 2;
 	public static final int THREE_CARDS = 3;
 	public static final int FOUR_CARDS = 4;
+	public static final int FIVE_CARDS = 5;
 
 	public static final int NUM_MILD_SHUFFLE_IN_GAME = 1;
+	static final int NUM_TARGETED_ATTACK_DRAW_COUNT = 2;
```
jazlynho2028: Is there a reason why only this variable is not public?

src/main/java/domain/Game.java
jazlynho2028: I suggested the loop so that if the constant NUM_TARGETED_ATTACK_DRAW_COUNT ever changes, you don't need to go through and change the implementation, BVA, and test cases

jazlynho2028: Thanks for making the other changes!

src/main/java/ui/PlayerDeckController.java
```diff
@@ -197,31 +219,29 @@ void onPlayCardsButton() {
...
-                case SEE_THE_FUTURE:
-                    view.buildSeeTheFutureOverlay(model.getSeeTheFutureCardIds());
+                    handleChangeCurrentPlayer(model.getCurrentPlayerIndex());
                     break;
                 case GODCAT:
                     view.bindGodcatConfirmButton(this::onGodcatConfirm);
                     view.buildGodcatOverlay(GameConstants.GODCAT_CARDTYPE_OPTIONS);
                     break;
+                case TARGETED_ATTACK:
+                    pendingTargetCard = cardType;
```
jazlynho2028: You can directly store the apply method as a Consumer<Integer> without going through the card type so that you don't need another switch statement

pendingTargetAction = model::applyTargetedAttack

src/main/java/ui/PlayerDeckController.java
```diff
@@ -15,14 +15,15 @@ public class PlayerDeckController {
 
     private final PlayerDeckView view;
     private final Game model;
+    CardType pendingTargetCard = null;
```
jazlynho2028: See R235 comment about changing this to 
Consumer<Integer> pendingTargetAction

As for the null problem, you can wrap it in Optional<>. Or you could do an empty lambda, but it gets a little more verbose when you have to reset the action to an empty lambda.

jazlynho2028 (changes requested):

src/main/java/domain/GameConstants.java
Luke-Meyer-hue: Sorry, I actually hadn't committed that one yet, I was planning to change it to a variable both targeted attack and attack could use but I hadn't gotten around to it. I just pushed it to be public static final int ATTACK_DRAW_COUNT

jazlynho2028: I see, that makes sense. Thanks!

src/main/java/ui/PlayerDeckView.java
```diff
@@ -936,4 +894,30 @@ public CardType getSelectedGodcatCardType() {
...
+        ObservableList<Node> nameTagButtons = playerNamesContainer.getChildren();
+
+        for (int i = 0; i < nameTagButtons.size(); i++) {
+            ToggleButton nameTagButton = (ToggleButton) nameTagButtons.get(i);
+            boolean isAtCurrentPlayerIndex = (i == currentPlayerIndex);
+
+            nameTagButton.setDisable(isAtCurrentPlayerIndex);
+
+            if (!isAtCurrentPlayerIndex) {
+                nameTagButton.setStyle("-fx-border-color: #ffcc00;" +
```
jazlynho2028: Don't control styling in java code. Edit existing .name-tag classes or add a new sub class to .name-tag if needed.

src/main/java/ui/PlayerDeckView.java
```diff
@@ -936,4 +894,30 @@ public CardType getSelectedGodcatCardType() {
...
+                        " -fx-border-width: 2px; -fx-cursor: hand;");
+            }
+        }
+    }
+
+    public void disableTargetSelectionMode(int currentPlayerIndex, boolean isGameOngoing) {
+        ObservableList<Node> nameTagButtons = playerNamesContainer.getChildren();
+
+        for (Node node : nameTagButtons) {
+            node.setStyle("");
```
jazlynho2028: Do node.getStyleClass().clear(); instead to match using css classes

src/main/java/ui/PlayerDeckController.java
```diff
@@ -83,7 +85,16 @@ public void setOnError(Consumer<String> handler) {
 
     void onNameTag(int playerIndex) {
```
jazlynho2028, line 86: Something like this would be better so that the controller doesn't rely on the current player button being disabled to determine logic. Also this way, you don't call the same handleChangeCurrentPlayer(playerIndex); twice. Note that updateTurnControls only calls a view method, so there's no harm in calling it in both cases.

```
void onNameTag(int playerIndex) {
        attempt(onError, () -> {
            if (model.getCurrentPlayerIndex() != playerIndex) {
                if (pendingTargetAction.isPresent()) {
                    pendingTargetAction.get().accept(playerIndex);
                    pendingTargetAction = Optional.empty();

                    view.disableTargetSelectionMode(model.getCurrentPlayerIndex(),
                            model.getIsGameOngoing());
                }

                handleChangeCurrentPlayer(playerIndex);
                updateTurnControls();
            }
        });
    }
```

src/main/java/ui/PlayerDeckView.java
```diff
@@ -936,4 +894,30 @@ public CardType getSelectedGodcatCardType() {
         return selectedGodcatCardType;
     }
 
+    public void enableTargetSelectionMode(int currentPlayerIndex) {
```
jazlynho2028: This method is almost identical to renderPlayerNameTags. You can just call renderPlayerNameTags(currentPlayerIndex, false) in here and change the styling in the css file if needed.

jazlynho2028: You might also want to change the isGameOngoing parameter name in renderPlayerNameTags to disableOtherPlayers to prevent misunderstanding. No need to change that parameter name outside of view though.

src/main/java/domain/Game.java
```diff
@@ -375,23 +340,28 @@ void applyClone() {
...
```
jazlynho2028: You should do a loop here to call turnManager.incrementTurn() until currentPlayerIndex == targetPlayerIndex. This is so that the roundCount gets updated properly. Keep in mind that you'll also need to check whether the target player is alive and throw if not once we have that addition pushed.

src/main/java/domain/Game.java
```diff
@@ -375,23 +340,28 @@ void applyClone() {
...
+        addAttackDrawCount();
+    }
+
+    void addAttackDrawCount() {
+        if (turnManager.getDrawCount() >= 2) {
+            turnManager.setDrawCount(
+                    turnManager.getDrawCount() + 2);
```
jazlynho2028: use GameConstants.ATTACK_DRAW_COUNT instead of 2 in turnManager.getDrawCount() + 2

src/main/java/domain/Game.java
```diff
@@ -267,75 +267,49 @@ public void advanceTurn() {
         }
         getCurrentPlayer().deselectHandCards();
         turnManager.incrementTurn();
+        turnManager.incrementDrawCount();
```
jazlynho2028, line 269: Because you removed draw count increment from incrementTurn, you need to add turnManager.incrementDrawCount() to all instances where incrementTurn() was called and update BVA and test cases accordingly. You did that here, but not for playExplode()

jazlynho2028 (changes requested):
Implementation is correct. Just a couple clean code modifications.

src/main/java/ui/PlayerDeckView.java
Luke-Meyer-hue: Ok, I'll add a .name-tag.targetable to the css file.

jazlynho2028: Is there a reason why you can't use the default .name-tag class? If you want the hand cursor you can just add that in .name-tag and make sure that .name-tag:selected, .name-tag:selected:disabled, and .name-tag:disabled have the normal arrow.

Luke-Meyer-hue: There isn't a reason that would really apply to here I suppose, the only reason I do it is because it helps me distinguish and if we ever wanted to separate them more its there. In this case though, I don't think it matters, so I can just add it to the .name-tag class

Luke-Meyer-hue: In fact, looking at both of the methods and renderPlayerNameTags, it actually looks as though both of the methods I added were unnecessary in the first place, I'm just going to remove them and solely use renderPlayerNameTags.

src/main/java/ui/PlayerDeckController.java
Luke-Meyer-hue: Alright, onNameTag should be updated as well as the tests and the BVA

src/main/java/domain/Game.java
Luke-Meyer-hue: I have added in the loop to keep track of roundCount, BVAs and tests have been adjusted as well

src/main/java/domain/Game.java
Luke-Meyer-hue: I have now added it to playExplode, however we'll have to remember to add this in for every other method that calls increment turn as well.

src/main/java/domain/Game.java
Luke-Meyer-hue: I have added it in to avoid random numbers in the code

docs/bva/TurnManager.md
```diff
@@ -95,3 +90,14 @@
...
   - **State of the system**: TurnManager constructed with 1 player, newPlayerIndex = 1
   - **Expected output**: throw IllegalArgumentException "error.invalidPlayerIndex"
+
+### Method under test: `incrementDrawCount()`
+- **TC14: Decrementing from zero throws state exception** ( :white_check_mark: )
+  - **Name of the test**: incrementDrawCount_validDrawCount_incrementedByOne
+  - **State of the system**: drawCount = 0
+  - **Expected output**: drawCount = 1
+
+- **TC15: incrementing from a positive value successfully increases count** ( :x: )
```
jazlynho2028: Mark this as done

src/main/java/ui/PlayerDeckController.java
```diff
@@ -197,31 +208,29 @@ void onPlayCardsButton() {
...
-                    break;
-                case SEE_THE_FUTURE:
-                    view.buildSeeTheFutureOverlay(model.getSeeTheFutureCardIds());
+                    handleChangeCurrentPlayer(model.getCurrentPlayerIndex());
                     break;
                 case GODCAT:
                     view.bindGodcatConfirmButton(this::onGodcatConfirm);
                     view.buildGodcatOverlay(GameConstants.GODCAT_CARDTYPE_OPTIONS);
                     break;
+                case TARGETED_ATTACK:
```
jazlynho2028: Make sure to pull from main to get the updated controller code for onPlayCardsButton. Godcat is intentionally separated from the other cards.

jazlynho2028 (approved):
Looks good. Just one minor bva update and some reminders about merging

PR #136, opened 2026-06-07
Wip catomic bomb revision

jazlynho2028:
Changed catomic bomb to advance turn like a skip instead of a super skip.


Luke-Meyer-hue (approved):
Looks good. It's a simple change so that instead of ending all the turns a person has stacked up it just ends one. Super Skip is the only card that should end all turns. Most changes are just the tests and BVA. Everything seems good.

PR #139, opened 2026-06-07
Card Implementation: Draw from Bottom

### LINKED ISSUE #114: Game Card-DrawFromTheBottom
Once played, this card goes to the discard pile. The user draws from the bottom of the deck instead of the top and their turn immediately ends.

Sunw00ha:


src/main/java/domain/Game.java
```diff
@@ -420,7 +387,15 @@ void applySwapTopAndBottom() {
     }
 
     void applyDrawFromTheBottom() {
```
jazlynho2028: Just like drawFromPile, you should check if the drawn card is an exploding kitten

jazlynho2028 (changes requested):
Needs significant change to account for exploding kitten draw and UI update. The below changes are applicable to other cards that involve automatic draw.


```
public Card applyDrawFromTheBottom() {
        Card card = drawPile.peekBottom();

        if (card.getType() != CardType.EXPLODING_KITTEN) {
            drawPile.removeBottom();
            getCurrentPlayer().addCardToHand(card);
        }

        turnManager.decrementDrawCount();

        return card;
    }
```

Because the code would be very similar between Game's drawFromPile and applyDrawFromTheBottom, you might want to make a private helper 

```
private Card drawCard(Supplier<Card> peek, Runnable remove) {
    Card card = peek.get();

    if (card.getType() != CardType.EXPLODING_KITTEN) {
        remove.run();
        getCurrentPlayer().addCardToHand(card);
    }

    turnManager.decrementDrawCount();
    return card;
}
```
Then drawFromPile and applyDrawFromTheBottom would be
```
public Card drawFromPile() {
        return drawCard(drawPile::peekTop, drawPile::removeTop);
    }

public Card applyDrawFromTheBottom() {
        return drawCard(drawPile::peekBottom, drawPile::removeBottom);
    }
```

You also need to update the controller just like after a normal draw. This is the current onDrawPile method in the controller.
```
void onDrawPile() {
        attempt(onError, () -> {
            Card drawnCard = model.drawFromPile();

            if (drawnCard.getType() == CardType.EXPLODING_KITTEN) {
                handleDrawExplodingKitten(drawnCard.getId());
            }
            else {
                rebindHandCards();
                updateDrawPile();
                updateTurnControls();
            }
        });
    }
```
Then you should define a helper 
```
private void handleDraw(Supplier<Card> draw) {
    attempt(onError, () -> {
        Card drawnCard = draw.get();

        if (drawnCard.getType() == CardType.EXPLODING_KITTEN) {
            handleDrawExplodingKitten(drawnCard.getId());
        }
        else {
            rebindHandCards();
            updateDrawPile();
            updateTurnControls();
        }
    });
}
```
So onDrawPile becomes

```
void onDrawPile() {
       handleDraw(model::drawFromPile);
}
```

You should edit updateByCardType (helper in onPlayCardsButton) to be

```
void updateByCardType(CardType cardType) {
        switch (cardType) {
            case SKIP:
            case SUPER_SKIP:
            case CATOMIC_BOMB:
                renderNextTurn();
                break;
            case SEE_THE_FUTURE:
                view.buildSeeTheFutureOverlay(model.getSeeTheFutureCardIds());
                break;
            case DRAW_FROM_THE_BOTTOM:
                handleDraw(model::applyDrawFromTheBottom);
                break;
            default:
                break;
        }
    }
```
Your branch is outdated right now. Make sure to pull from main so that you have the updated game and controller methods.

src/main/java/domain/Game.java
Sunw00ha: Done: updated Draw From The Bottom to use the shared draw helper so it now handles Exploding Kittens the same way as a normal draw. Also updated the controller to route it through the normal draw UI path, and updated the related tests/BVA.

src/main/java/domain/Game.java
```diff
@@ -269,15 +266,12 @@ public Card drawFromPile() {
...
-            getCurrentPlayer().addCardToHand(card);
-        }
-        turnManager.decrementDrawCount();
-        getCurrentPlayer().deselectHandCards();
-        return card;
+    public Card drawFromPile() {
+        return drawCard(drawPile::peekTop, drawPile::removeTop);
+    }
+
+    public Card applyDrawFromTheBottom() {
```
jazlynho2028: can you rename this to drawFromTheBottom with DJ's naming for drawRecycle (including BVA and test instances of the name)? It's good to separate public non void methods with private void ones this way too.

src/main/java/ui/PlayerDeckController.java
```diff
@@ -147,12 +148,25 @@ void onNameTag(int playerIndex) {
         });
     }
 
-    void onDrawPile() {
+    private void handleDraw(Supplier<Card> draw) {
```
jazlynho2028: Since DJ already made a handleDrawnCard method that basically does the same thing except it takes in a Card instead of a Suppler<Card>, you can just use handleDrawnCard(model.drawFromTheBottom()) and change onDrawPile back to handleDrawnCard(model.drawFromPile()). DJ's way of doing it is cleaner than the one I suggested last week since it passes an object rather than a method.

src/main/java/domain/Game.java
jazlynho2028: can you also move this method and its BVA and test case order according to the CardType enum order like the rest of the card-specific methods?

jazlynho2028 (changes requested):
Looks good overall! Just some minor changes to make sure the implementation is consistent for all normal draws + card induced draws.

src/main/java/ui/PlayerDeckController.java
jazlynho2028: The Game helper method is okay to pass in methods though since we need to run peek.get() and remove.run() conditionally

src/main/java/domain/Game.java
Sunw00ha: done!

src/main/java/ui/PlayerDeckController.java
Sunw00ha: Done!

jazlynho2028 (approved):
Looks great!

PR #141, opened 2026-06-09
Card Implementation: Ragebait

### LINKED ISSUE #117: Game OriginalCard-Ragebait
When used, this card goes to the discard pile. The user swaps their hand with a targeted player.

Luke-Meyer-hue:
When played, the user can swap cards with a targeted player. The user's turn continues as normal. 

Adjusted onNameTag to accept a lambda because Ragebait doesn't automatically end the turn, and I didn't want to keep adding on conditionals or new methods when a lambda is simpler.


src/main/java/ui/PlayerDeckController.java
```diff
@@ -229,20 +232,34 @@ void onPlayCardsButton() {
...
+            case SUPER_SKIP:
+            case CATOMIC_BOMB:
+                renderNextTurn();
                 break;
             case SEE_THE_FUTURE:
                 view.buildSeeTheFutureOverlay(model.getSeeTheFutureCardIds());
                 break;
             case TARGETED_ATTACK:
-                pendingTargetAction = Optional.of(model::applyTargetedAttack);
+                pendingTargetAction = Optional.of(targetIndex -> {
```
jazlynho2028: in all cases (targeted attack, ragebait, player select before game starts), updateTurnControls can/should be called. So no need to put that in the pending target lambda. Just do it after the if block in onNameTags

src/main/java/ui/PlayerDeckController.java
```diff
@@ -229,20 +232,34 @@ void onPlayCardsButton() {
...
-			case CATOMIC_BOMB:
-				renderNextTurn();
+            case SUPER_SKIP:
+            case CATOMIC_BOMB:
+                renderNextTurn();
                 break;
             case SEE_THE_FUTURE:
                 view.buildSeeTheFutureOverlay(model.getSeeTheFutureCardIds());
                 break;
             case TARGETED_ATTACK:
```
jazlynho2028, line 243: This would be easier to read and test if you put the calls in pending in a helper for targeted attack and ragebait

jazlynho2028 (changes requested):

src/main/java/ui/PlayerDeckController.java
Luke-Meyer-hue: Thanks for that, I wasn't sure if there were any special cases. I'll move it over and adjust the BVA/Tests.

src/main/java/ui/PlayerDeckController.java
Luke-Meyer-hue: Sounds good, I'll simply make the calls helper methods

jazlynho2028: Note that when you're in player select mode, players shouldn't be allowed to press anything other than the name tags to prevent unexpected behavior.

```
case TARGETED_ATTACK:
        pendingTargetAction = Optional.of(model::applyTargetedAttack);
        enablePlayerSelect();
        break;

private void enablePlayerSelect() {
        enableNameTags();
        disableAllButNameTags();
    }

private void enableNameTags() {
    view.renderPlayerNameTags(model.getCurrentPlayerIndex(), true,
            model.getAliveIndices());
}

private void disableAllButNameTags() {    
    view.renderDrawPile(false, model.isDrawPileEmpty());
    view.renderHandVisibilityButton(model.getIsFaceUp(), false);
    view.buildAndAddPlayerHandCards(
            model.getCurrentPlayerHandIds(), model.getIsFaceUp(), false
    );
    view.renderTurnControlSection(false, false);
}
```

jazlynho2028 (changes requested):
Mostly looks good. Just one more point about enabling player select

jazlynho2028 (changes requested):
Is it just me or did you put the new player method BVAs and tests in the TurnManager files

Luke-Meyer-hue:
> Is it just me or did you put the new player method BVAs and tests in the TurnManager files

Nope, you are right, I did actually do that, I don't know why but I get them mixed up a lot

Luke-Meyer-hue:
Moved swapHands tests over to Player.md and PlayerTests, they were misplaced

jazlynho2028 (approved):
Looks good now, thanks!

Luke-Meyer-hue:
Thanks for the reviews!

PR #145, opened 2026-06-12
Card Implementation: Mild Shuffle

### LINKED ISSUE #120: Game OriginalCard-MildShuffle
Once played, this card goes to the discard pile. The user shuffles the top three cards of the deck.

Sunw00ha:


src/main/java/domain/Deck.java
```diff
@@ -35,6 +35,34 @@ public void shuffle() {
...
+        if (numCards <= 1) {
+            return;
+        }
+
+        List<Card> cards = new ArrayList<>(deck);
+        List<Card> topCards = new ArrayList<>(cards.subList(0, numCards));
+        List<Card> remainingCards = new ArrayList<>(
+                cards.subList(numCards, cards.size()));
+
+        for (int i = topCards.size() - 1; i > 0; i--) {
```
jazlynho2028, line 44: This block of code is the same as the one in shuffle(). You can either make a private helper and call it in shuffle() and shuffleTopNCards or just call one line shuffleTopNCards(deck.size()) in shuffle.

src/main/java/domain/Game.java
```diff
@@ -518,7 +518,7 @@ void applyDoubleUp() {
     }
 
     void applyMildShuffle() {
-        // TODO
+        drawPile.shuffleTopNCards(GameConstants.THREE_CARDS);
```
jazlynho2028: You should define a GameConstants.MILD_SHUFFLE_SHUFFLE_COUNT instead

jazlynho2028 (changes requested):
Minor clean code changes. Otherwise looks good!

src/main/java/domain/Deck.java
Sunw00ha: Done!

src/main/java/domain/Game.java
Sunw00ha: Done!

jazlynho2028 (approved):
Looks good!

PR #147, opened 2026-06-12
Card implementation: Clone Card

### LINKED ISSUE #112: Game Card-Clone
When played, this card goes to the discard pile. The user then plays the card under the Clone card. In other words, once played, the user plays the top card of the discard pile.

Sunw00ha:


src/main/java/domain/Game.java
```diff
@@ -450,8 +449,20 @@ public void applyGodcat(CardType cardType) {
         }
     }
 
-    void applyClone() {
-        // TODO
+    CardType applyClone() {
+        List<Card> topDiscardCards = discardPile.peekTopNCards(
+                GameConstants.CLONE_PEEK_COUNT);
+
+        if (topDiscardCards.size() < GameConstants.CLONE_PEEK_COUNT) {
```
jazlynho2028: You don't need this check anymore. Since the game only supports plays for one card at a time (except for special cases for two and three of a kind), clone wouldn't ever "clone" more than one card at a time. You can remove the CLONE_PEEK_COUNT constant from GameConstants.

src/main/java/domain/Game.java
```diff
@@ -450,8 +449,20 @@ public void applyGodcat(CardType cardType) {
...
+            throw new IllegalStateException("error.noCardToClone");
+        }
+
+        Card clonedCard = topDiscardCards.get(GameConstants.CLONE_CARD_INDEX);
```
jazlynho2028: You don't need CLONE_CARD_INDEX or this line anymore since peekTop will give you the clonedCard already

src/main/java/domain/Game.java
```diff
@@ -450,8 +449,20 @@ public void applyGodcat(CardType cardType) {
         }
     }
 
-    void applyClone() {
-        // TODO
+    CardType applyClone() {
```
jazlynho2028, line 452: I like that you have applyClone return the CardType directly so that the controller doesn't need extra management for clone unlike godcat

src/main/java/domain/Game.java
```diff
@@ -450,8 +449,20 @@ public void applyGodcat(CardType cardType) {
...
+        List<Card> topDiscardCards = discardPile.peekTopNCards(
```
jazlynho2028: You can use discardPile.peekTop(), which already checks that the deck is not empty, so you don't need to handle that exception case.

jazlynho2028 (changes requested):
Looks good but can be simplified.

src/main/java/domain/Game.java
Sunw00ha: Done!

src/main/java/domain/Game.java
Sunw00ha: Done!

src/main/java/domain/Game.java
Sunw00ha: Done!

src/main/java/domain/Game.java
```diff
@@ -150,19 +150,26 @@ public CardType playSelectedCards() {
         List<Card> selectedCards = getCurrentPlayer().getSelectedCards();
         CardType cardType = selectedCards.get(0).getType();
 
+        Card clonedCard = null;
```
jazlynho2028: Is there a reason why you didn't keep applyClone as this?

```
CardType applyClone() {
   CardType clonedCardType = discardPile.peekTop().getType();
   applyByCardType(cloneCardType);
   
   return clonedCardType;
}
```

Clean code says that null shouldn't be passed around.

src/main/java/domain/Game.java
```diff
@@ -150,19 +150,26 @@ public CardType playSelectedCards() {
...
+
         for (Card card : selectedCards) {
             card.toggleSelected();
 
             getCurrentPlayer().removeCardFromHand(card);
             discardPile.addCardToTop(card);
         }
 
-        applyByCardType(cardType);
+        if (cardType == CardType.CLONE) {
```
jazlynho2028: Why do you have a check to return applyClone here and another a return applyClone call in applyByCardType?

jazlynho2028 (changes requested):
I thought your original implementation of clone was cleaner. Was there a reason why you ended up doing it this way? You're essentially calling/implementing clone effects in three separate places here.

docs/bva/Game.md
```diff
@@ -1726,3 +1727,16 @@
...
+
+### Method under test: `applyClone()`
+- **TC226: Apply Clone with Attack under Clone** ( :white_check_mark: )
+  - **Name of the test**: applyClone_attackUnderClone_appliesAttackAndReturnsAttack
+  - **State of the system**: discardPile.peekTopNCards(GameConstants.CLONE_PEEK_COUNT) returns [CLONE, ATTACK]
+  - **Expected output**:
+    - applyAttack is called
+    - returns CardType.ATTACK
+
+- **TC227: Apply Clone with See The Future under Clone** ( :white_check_mark: )
```
jazlynho2028, line 1739: Might be useful to add a test case where there's a Skip (or any other card) under a Clone that's under another Clone. The expected behavior is that the top Clone will activate the second Clone, which will activate the card below it (Skip).

jazlynho2028 (changes requested):

src/main/java/domain/Game.java
Sunw00ha: Good point: I originally changed it to address the earlier peekTop() suggestion, but I agree that the updated version split Clone logic across too many places. I reverted closer to the original structure so Clone handling stays centralized in applyClone(), and removed the null-passing approach.

src/main/java/domain/Game.java
Sunw00ha: Addressed in the comment above! I reverted it so Clone is no longer handled separately in playSelectedCards() and is only handled through applyByCardType().

docs/bva/Game.md
Sunw00ha: Good idea: I added a nested Clone test where Clone copies another Clone, which then applies the card underneath it.

jazlynho2028 (approved):
Looks good now!

PR #148, opened 2026-06-13
Integration Testing: Catomic Bomv

Luke-Meyer-hue:
These are integration tests using applyCatomicBomb as the basis.


dongjoo224 (approved):
Tests all look good to me. The scenarios taken into account make sense and the BVA looks good.

PR #152, opened 2026-06-15
Card Implementation: Feral Cat

### LINKED ISSUE #107: Game Card-FeralCat
This card is able to be used as any cat card, so they can be a pair or three of a kind with any other one card or pair. If used as one of a pair, the user gets to steal a random card from a player of choice. If used as one of three of a kind, the user gets to name a specific card, and if the targeted player has it, they must give it to the person that played the three of a kind.

ejeankuo:


Luke-Meyer-hue (approved):
It seems good to me, adding in not only the Feral cat, but allowing the Godcat card to be used as a cat card and the clone card as well if it applies

